PR #29, opened 2019-03-07
Sandbox build initial release

luigibk:
### Change description ###

```
[ ] Yes
[x] No
```


k8s/sandbox/sealed-secrets/sealed-secrets.yaml
```diff
@@ -0,0 +1,18 @@
...
+  releaseName: sealed-secrets
+  chart:
+    repository: https://kubernetes-charts.storage.googleapis.com/
+    name: sealed-secrets
+    version: 1.0.1
+  values:
+    image:
+      repository: quay.io/bitnami/sealed-secrets-controller
+      tag: v0.7.0
+    secretName: "sealed-secrets-key"
```
timwebster9: I think for preview/aat I used an explicitly different key here because the controller also generates one with the same name (but I think only if there isn't one already there).  the point being if something doesn't work it's hard to know which key it's using

luigibk: Good point. Thanks

PR #35, opened 2019-03-08
Adding config for ccd-case-emanagement-web api, ccd-api-gateway-web

adusumillipraveen:
Adding config for ccd-case-emanagement-web api

https://tools.hmcts.net/jira/browse/RPE-1005

### Change description ###

Adding config for ccd-case-emanagement-web api, ccd-api-gateway-web

```
[ ] Yes
[x] No
```


k8s/aat/ccd/ccd-case-management-web/ccd-case-management-web.yaml
```diff
@@ -0,0 +1,26 @@
...
+  chart:
+    repository: https://hmcts.azurecr.io/helm/v1/repo/
+    name: ccd-case-management-web
+    version: 0.0.2
+  values:
+    nodejs:
+      replicas: 2
+      useInterpodAntiAffinity: true
+      applicationInsightsInstrumentKey: "f61ab34c-07da-42fc-ac7e-3f8cdadfd5a1"
+      image: hmcts.azurecr.io/hmcts/ccd-case-management-web:latest
```
timja: don't you need some environment config like where the api gateway is?

adusumillipraveen: They are all in https://github.com/hmcts/ccd-case-management-web/blob/master/charts/ccd-case-management-web/values.aat.template.yaml

timja: two problems there,
1. values.aat.template isn't published / used, we should probably get rid of it as flux is replacing that
2. those values need to be different for new AAT, i.e. using the internal urls for those apps

PR #86, opened 2019-04-25
Update kured secrets for kured chart 0.0.4

luigibk:

https://tools.hmcts.net/jira/browse/RPE-1090

### Change description ###

```
[ ] Yes
[x] No
```


luigibk:
> I guess you are going to stick to using SSH to clone the chart instead of using the new published one? I don't have a preference just wondering if you think it's better one way or the other

I'd rather use the stable chart to be honest, but after having a look I can say that it cannot work for us unless we make same (minor I think) changes. Until then the ssh clone works

timwebster9:
> > I guess you are going to stick to using SSH to clone the chart instead of using the new published one? I don't have a preference just wondering if you think it's better one way or the other
> 
> I'd rather use the stable chart to be honest, but after having a look I can say that it cannot work for us unless we make same (minor I think) changes. Until then the ssh clone works

I meant the chart we publish ourselves... with Flux you can either SSH clone it or use the one published in ACR.  Currently we're SSH cloning it, but with your changes in the other PR we'll have a newer version, so could just use that instead

PR #170, opened 2019-06-04
Papi sbox

tkimber-hmcts:

https://tools.hmcts.net/jira/browse/RDO-3707

### Change description ###

Added 'papi-sandbox' Flux configuration.

```
[ ] Yes
[x] No
```


k8s/sandbox/papi-cluster-00/flux/flux.yaml
```diff
@@ -0,0 +1,34 @@
+---
```
timja: this is in the wrong folder should be k8s/papi-sandbox

k8s/sandbox/papi-common/papi/namespace.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+   name: papi-sbox
```
timja: having a namespace after a cluster is weird is it not multiple teams using it?
anyway, the env shouldn't be in it.
```suggestion
   name: papi
```

k8s/sandbox/papi-common/papi/identity.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: "aadpodidentity.k8s.io/v1"
+kind: AzureIdentity
+metadata:
+ name: papi-sbox
```
timja: i would probably drop the env throughout this file except in the resource id

PR #198, opened 2019-06-18
Add config for sandbox management

adusumillipraveen:
Fixes [AB#214](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/214)

### Change description ###

Overriding the old mgmt sandbox cluster folder . But nothing looks like to break the old one.

```
[ ] Yes
[ ] No
```


k8s/mgmt-sandbox/cluster-00/traefik/traefik.yaml
```diff
@@ -0,0 +1,39 @@
...
+        cpu: 1000m
+        memory: 300Mi
+    service:
+      annotations:
+        service.beta.kubernetes.io/azure-load-balancer-internal: "true"
+    rbac:
+      enabled: true
+    dashboard:
+      enabled: true
+      domain: traefik00.service.core-compute-mgmt-sandbox.internal
```
timja: management doesn't use these domains (no ASE network and that's what core-compute is)
historically we've used
.platform.hmcts.net with a sandbox prefix in the domain name
there's also the option of doing .sandbox.platform.hmcts.net

PR #199, opened 2019-06-18
Add camunda helm release

njrich28:


k8s/sandbox/common/camunda/camunda-bpm.yaml
```diff
@@ -0,0 +1,44 @@
...
+    livenessFailureThreshold: 3
+    replicas: 1
+    aadIdentityName: camunda
+    databaseType: external   # external or pod. Note: for pod the 'postgresql-pod' tag below needs to be true
+    tags:
+      postgresql-pod: false
+    postgresqlServerName: "camunda-db-sandbox"
+    postgresqlHost: "camunda-db-sandbox.postgres.database.azure.com"
+    postgresqlDatabase: "camunda"
+    resourceGroup: camunda-sandbox   # osba only
```
timja: why osba only?

k8s/sandbox/common/camunda/camunda-bpm.yaml
```diff
@@ -0,0 +1,44 @@
...
+  namespace: camunda
+  annotations:
+    flux.weave.works/automated: "false"
+spec:
+  releaseName: camunda-bpm
+  chart:
+    repository: https://hmcts.azurecr.io/helm/v1/repo/
+    name: camunda
+    version: 1.1.0
+  values:
```
timja: most of these values should be set in the chart.
probe, application port etc...

k8s/sandbox/common/camunda/camunda-bpm.yaml
njrich28: cut-n-paste relic - will check

PR #219, opened 2019-06-21
Adding kustomize to reduce duplications across environments

adusumillipraveen:
Adding Kustomize to flux , to be able to override base manifests at env level instead of duplicating them across. Testing only on sandbox 01 for now.

Working on [AB#224](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/224)

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/common-base/traefik/traefik.yaml
+++ b/k8s/common-base/traefik/traefik.yaml
@@ -1,0 +1,34 @@
+apiVersion: flux.weave.works/v1beta1
+kind: HelmRelease
+metadata:
+  name: traefik
+  namespace: admin
+  annotations:
+    flux.weave.works/automated: "false"
+spec:
+  releaseName: traefik
+  chart:
+    repository: https://kubernetes-charts.storage.googleapis.com/
+    name: traefik
+    version: 1.61.1
+  values:
+    imageTag: 1.7.9
+    resources:
+      requests:
+        cpu: 100m
+        memory: 20Mi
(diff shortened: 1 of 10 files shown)


k8s/common-base/traefik/traefik.yaml
```diff
@@ -0,0 +1,34 @@
...
+    resources:
+      requests:
+        cpu: 100m
+        memory: 20Mi
+      limits:
+        cpu: 1000m
+        memory: 300Mi
+    service:
+      annotations:
+        service.beta.kubernetes.io/azure-load-balancer-internal: "true"
```
timja: hmm I wonder if we can override this to empty in the customisation?
something like:
```
service:
  annotations: []
```
as demo uses a public load balancer

k8s/common-base/traefik/traefik.yaml
```diff
@@ -0,0 +1,34 @@
...
+    rbac:
+      enabled: true
+    dashboard:
+      enabled: true
+    ssl:
```
timja: may as well set the ssl ciphers etc globally as well

k8s/common-base/traefik/traefik.yaml
adusumillipraveen: Yes, tested in local , it works. Flux should just apply the result of Kustomise.

PR #242, opened 2019-06-25
CMC: add mailhog deployment to demo/money-claims

markry-kainos:
### Change description ###

no peering for `mta.reform.hmcts.net` but probably don't want traffic to go there anyway 🤔 this follows on from desired service + all dependencies. we will share between all our demo instances.

could add with ingress if needed - but can port-forward for now.

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/demo/common/money-claims/mailhog/mailhog.yaml
+++ b/k8s/demo/common/money-claims/mailhog/mailhog.yaml
@@ -36,7 +36,7 @@ metadata:
     app: mailhog
   annotations:
     flux.weave.works/automated: "true"
-    flux.weave.works/tag.mailhog: regexp:^1.*
+    flux.weave.works/tag.mailhog: regexp:^v1.*
 spec:
   template:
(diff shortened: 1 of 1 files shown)


k8s/demo/common/money-claims/mailhog/mailhog.yaml
```diff
@@ -0,0 +1,73 @@
...
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: mailhog
+  namespace: money-claims
+  labels:
+    app: mailhog
+  annotations:
+    flux.weave.works/automated: "true"
+    flux.weave.works/tag.mailhog: regexp:^v1.*
```
markry-kainos: dont think semver will work with the `v` 🤔

timja: dunno, I wouldn't worry too much

PR #325, opened 2019-07-18
Enable prometheus integration with Azure monitor

timja:
(sandbox only for now)


adusumillipraveen:
Awesome, finally its out at the right time :)

PR #342, opened 2019-07-31
Add CODEOWNERS

timja:
[AB#277](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/277)

### Change description ###
Add CODEOWNERS to restrict who can review and merge

```
[ ] Yes
[ ] No
```


CODEOWNERS
```diff
@@ -0,0 +1,30 @@
...
+CODEOWNERS @hmcts/platform-engineering @hmcts/devops
+
+# AAT applications
+## environment rules
+k8s/aat/ @hmcts/platform-engineering @hmcts/devops
+
+## application rules
+
+### RPE
+k8s/aat/common/rpe/namespace.yaml @hmcts/platform-engineering
```
adusumillipraveen: I thought we should be able to do k8s/**/common/rpe/namespace.yaml @hmcts/platform-engineering. so that we can avoid duplication.

timja:
image

timja:
image

CODEOWNERS
timja: not sure if we really get any benefit with that and we lose explicitness. don't really mind though

adusumillipraveen: I think it will be good to reduce duplication. but it would mean k8s/sandbox will also have same approvals . But worried about this file growing too big.

timja: let's give it a go have pushed a commit

timja:
@noferrets 
> Can we move k8s/prod line to the last line. This will ensure it cannot be accidentally overridden.
https://help.github.com/en/articles/about-code-owners
```
# Order is important; the last matching pattern takes the most
# precedence. When someone opens a pull request that only
# modifies JS files, only @js-owner and not the global
# owner(s) will be requested for a review.
```

Least specific first

I've added the test you requested and I've validated it works, screenshots + example broken builds are on this PR

PR #352, opened 2019-08-02
config-map branch does not exist

naikajah:
Config-map branch does not exist in the helm neuvector https://github.com/hmcts/neuvector-helm

```
[ ] Yes
[X] No
```


k8s/common-base/neuvector/neuvector.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   timeout: 600
   chart:
     git: https://github.com/hmcts/neuvector-helm
```
timja: we can use upstream now, as they merged our change

PR #401, opened 2019-08-12
Fix neuvector config

timja:


timja:
Overriding codeowners due to Sev 0 alerts spamming our inboxes every 5 minutes.
For everyones sanity and so important emails don't get missed...

Will add these apps to codeowners tomorrow.

PR #413, opened 2019-08-15
ccd-demo integrate components

akdasari:
### Change description ###
added payment-api, draft-store-service, case-manegement-web

```
[ ] Yes
[x ] No
```

### DIFF
--- a/k8s/demo/common/ccd/ccd-demo.yaml
+++ b/k8s/demo/common/ccd/ccd-demo.yaml
@@ -35,12 +35,23 @@ spec:
           secretKeyRefName: ccd-admin-web-oauth2-client-secret
           secretKeyRefKey: key
     apiGateway:
-      environemnt:
+      environment:
         s2sKey: AAAAAAAAAAAAAAAA
         idamClientSecret:
           fromSecret: true
           secretKeyRefName: ccd-api-gateway-oauth2-client-secret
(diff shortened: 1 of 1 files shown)


k8s/demo/common/ccd/ccd-demo.yaml
```diff
@@ -52,12 +63,47 @@ spec:
...
+        releaseNameOverride: ccd-demo-draftstore
+        environment:
+          IDAM_URL: https://idam-api.demo.platform.hmcts.net
+          S2S_URL: http://ccd-demo-s2s
+          DRAFT_STORE_DB_HOST: ccd-demo-postgres
     rpe-service-auth-provider:
       java:
         releaseNameOverride: ccd-demo-s2s
+
+    payment-api:
```
timja: Where's the corresponding change to chart-ccd to add this?
I would expect to see the other services here:
https://github.com/hmcts/chart-ccd/blob/master/ccd/requirements.yaml

https://github.com/hmcts/chart-cmc/blob/master/cmc/requirements.yaml#L25

akdasari: we have paymentsUrl in Values.yaml  and in turn paymentsUrl is used in api-gateway.yaml template. hence i added this  payment-api:

PR #454, opened 2019-08-19
Em 1936 add charts for demo

yogesh-hullatti:
https://tools.hmcts.net/jira/browse/EM-1936

Add in the required files for EM-Annotation module


k8s/demo/common/em-annotation/em-annotation-demo.yaml
```diff
@@ -0,0 +1,57 @@
...
+    enable: true
+    force: true
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: em-annotation
+    version: 0.0.1
+  values:
+    ### EM-ANNOTATION SERVICES ###
+    em-anno:
+      releaseNameOverride: em-annotation-em-anno
```
markry-kainos: you could drop the `em-annotation-` from all these but its upto you

PR #473, opened 2019-08-21
Added in the files for dg-docassembly

yogesh-hullatti:
https://tools.hmcts.net/jira/browse/EM-1978

Added in the required files

### DIFF
--- a/k8s/demo/common/dg-docassembly/dg-docassembly-demo.yaml
+++ b/k8s/demo/common/dg-docassembly/dg-docassembly-demo.yaml
@@ -1,0 +1,50 @@
+---
+apiVersion: flux.weave.works/v1beta1
+kind: HelmRelease
+metadata:
+  name: dg-docassembly-demo
+  namespace: dg-docassembly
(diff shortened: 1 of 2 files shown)


k8s/demo/common/dg-docassembly/dg-docassembly-demo.yaml
```diff
@@ -0,0 +1,50 @@
...
+  releaseName: dg-docassembly-demo
+  timeout: 900
+  forceUpgrade: true
+  rollback:
+    enable: true
+    force: true
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: dg-docassembly
+    version: 0.0.1
```
markry-kainos: is this correct? 🤔 just think i remember a bigger num

yogesh-hullatti: Not sure. Just started with this number as this was the first version of the file

PR #650, opened 2019-09-19
Prod config for Ethos

jacarrete:
### Change description ###

Added Ethos prod config

```
[ ] Yes
[ X] No
```

### DIFF
--- a/k8s/prod/common/ethos/repl-docmosis-service.yaml
+++ b/k8s/prod/common/ethos/repl-docmosis-service.yaml
@@ -30,6 +30,8 @@ spec:
       image: hmctspublic.azurecr.io/ethos/repl-docmosis-backend:prod-ec0fade0
       ingressHost: ethos-repl-docmosis-backend-prod.service.core-compute-prod.internal
       environment:
+        TORNADO_URL: https://docmosis.platform.hmcts.net/rs/render
+        IDAM_API_URL: https://idam-api.platform.hmcts.net
         JAVA_TOOL_OPTIONS: -XX:InitialRAMPercentage=30.0 -XX:MaxRAMPercentage=65.0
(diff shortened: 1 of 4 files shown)


k8s/prod/common/ethos/repl-docmosis-service.yaml
```diff
@@ -0,0 +1,41 @@
...
+      replicas: 2
+      memoryRequests: '512Mi'
+      cpuRequests: '250m'
+      memoryLimits: '2048Mi'
+      cpuLimits: '1500m'
+      readinessDelay: 45
+      readinessTimeout: 5
+      readinessPeriod: 15
+      useInterpodAntiAffinity: true
+      applicationInsightsInstrumentKey: "162CD82F0363F36E81218AD218C3C19F862CD480"
```
adusumillipraveen: Is that right app insights key ?

jacarrete: Yes, I got it from the prod secrets. At some point it should be removed and created automatically as I don't think we are using App Insights properly.

k8s/prod/common/ethos/repl-docmosis-service.yaml
```diff
@@ -0,0 +1,41 @@
...
+      image: hmctspublic.azurecr.io/ethos/repl-docmosis-backend:prod-ec0fade0
+      ingressHost: ethos-repl-docmosis-backend-prod.service.core-compute-prod.internal
+      environment:
```
timja: It looks like these 2 values need overriding for production:
https://github.com/hmcts/ethos-repl-docmosis-service/blob/master/charts/ethos-repl-docmosis-backend/values.yaml#L23-L24

jacarrete: Added

PR #662, opened 2019-09-20
Test if reinstalling fixes issue on prod01 with kured

adusumillipraveen:

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/prod/cluster-01-overlay/kustomization.yaml
+++ b/k8s/prod/cluster-01-overlay/kustomization.yaml
@@ -5,7 +5,6 @@ bases:
   - ../../admin/fluxcloud
   - ../../admin/kube-slack
   - ../../admin/traefik
-  - ../../kured
   - ../../neuvector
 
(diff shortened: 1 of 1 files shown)


k8s/prod/cluster-01-overlay/kustomization.yaml
```diff
@@ -22,9 +21,6 @@ patchesStrategicMerge:
   #traefik patch
   - ../../admin/traefik/patches/prod/cluster-01/traefik.yaml
 
-  #kured patch
-  - ../../kured/patches/prod/cluster-01/kured.yaml
```
timja: doubt this will fix it, I think there's an issue with the node, I've deleted the pod twice  and it's always the same node having the issue: aks-nodepool-19711934-1

adusumillipraveen: Ahhk, thought i would do default trouble shoot :D

PR #753, opened 2019-09-27
Use latest TYA Notif image

chris-moreton:
https://tools.hmcts.net/jira/browse/SSCS-6617

I think in the original PR for the creation of TYA Notif, I didn't use the "latest" version, instead commiting a tag reference which I'm guessing I pulled from Azure.

### DIFF
--- a/k8s/aat/common/sscs/sscs-tya-notif.yml
+++ b/k8s/aat/common/sscs/sscs-tya-notif.yml
@@ -22,7 +22,7 @@ spec:
       cpuLimits: '1500m'
       useInterpodAntiAffinity: true
       applicationInsightsInstrumentKey: "e2300a5f-a16f-4657-a63b-41913805661e"
-      image: hmctspublic.azurecr.io/sscs/tya-notif:prod-e07d0898
+      image: hmctspublic.azurecr.io/sscs/tya-notif:latest
       aadIdentityName: sscs
     global:
       environment: aat


k8s/aat/common/sscs/sscs-tya-notif.yml
```diff
@@ -22,7 +22,7 @@ spec:
       cpuLimits: '1500m'
       useInterpodAntiAffinity: true
       applicationInsightsInstrumentKey: "e2300a5f-a16f-4657-a63b-41913805661e"
-      image: hmctspublic.azurecr.io/sscs/tya-notif:prod-e07d0898
```
adusumillipraveen: not sure why u want to do this ? flux will manage tags based on https://github.com/hmcts/cnp-flux-config/pull/753/files#diff-4ab970973ca77db6a0ac8017993800d0L9

chris-moreton: Thanks @adusumillipraveen - @jackmaloney mentioned that we might need to set it to "latest" the first time around, but if Flux will manage it in any case, that's great. Was just hoping it might be the reason we can't get it to deploy :(

adusumillipraveen: Shouldn't be the case.

timja: you can commit latest in the first commit, and then flux manages it for you from then on

PR #785, opened 2019-10-02
Attempting the equalise properties with AAT environments

jvictorcf:
### Change description ###
Divorce data extraction e-mails are not being sent from Prod since COS was migrated to AKS.
This is an attempt to equalise all e-mail related properties and see if it starts working.

```
[ ] Yes
[X ] No
```


k8s/prod/common/divorce/div-cos.yaml
```diff
@@ -31,6 +31,7 @@ spec:
       environment:
         MANAGEMENT_ENDPOINT_HEALTH_CACHE_TIMETOLIVE: "4000"
         IDAM_API_URL: "https://idam-api.platform.hmcts.net"
+        SPRING_MAIL_PROPERTIES_MAIL_SMTP_STARTTLS_ENABLE: false
```
hosainnet: This was true though https://github.com/hmcts/div-case-orchestration-service/blob/master/infrastructure/variables.tf#L139 (I cannot see it overridden in prod in *.tf)

jvictorcf: That's true. The thing is that it's the only difference I can see between AAT (which works) and PROD (which doesn't)

hosainnet: Okay worth a try I guess..

PR #877, opened 2019-10-11
Add cert + require ssl for new traefik on aat 01

luigibk:

[[AB#597](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/597)]

### Change description ###

```
[ ] Yes
[x] No
```


k8s/aat/common/traefik/traefik-neuvector-api-values.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: bitnami.com/v1alpha1
+kind: SealedSecret
+metadata:
+  creationTimestamp: null
+  name: traefik-neuvector-api-values
```
timja: is this specific to the neuvector api at all?
could it just be traefik-cert-values

luigibk: The traefik instance is named traefik-neuvector-api.

timja: another traefik could use the cert though without having to store the cert in the repo twice

PR #909, opened 2019-10-15
Added in AppInsights key

yogesh-hullatti:
Added in AppInsights key


yogesh-hullatti:
This change is no more required as Luigi managed to add the appInsightsKey for Prod manually on to azure portal

PR #946, opened 2019-10-18
Revert "Temp remove notify config (#923)"

jackmaloney:
This reverts commit 6f5fe4715152478e709ace6560b00e056688619c.


k8s/aat/common/sscs/sscs-tya-notif.yml
```diff
@@ -0,0 +1,27 @@
+---
```
timja: this must be missing in codeowners btw...
SSCS should be getting review requests...

jackmaloney: It is in there @timja: https://github.com/hmcts/cnp-flux-config/blob/tribunals-flux-config/CODEOWNERS#L112. Maybe because that is yml and not yaml as stated in code owners

PR #1014, opened 2019-10-23
Rdm 6194 -Added Print and case activity services

akdasari:

```
[ ] Yes
[ ] No
```


k8s/demo/common/ccd/latest-ccd-chart.yaml
```diff
@@ -74,14 +74,25 @@ spec:
...
-      enabled: true
-      nodejs:
-        ingressClass: 
-        ingressHost: 
+    
+    # CCD Supporting services
+    ccd:
+      printService:
+        enabled: true
+        nodejs:
```
mario-paniccia: 1) are you sure having `nodejs:` nested here will be picked up? I don't think
2) much of this info is already present in the chart-ccd. Ideally we should add the config there rather than here. If you want to add that here to test it more quickly, please add a TODO comment where you mention to move all this to the chart-ccd once is proved working.
Remember we want as little config as possible in flux

akdasari: it is right config Mario, as those url change in demo, hence override.

PR #1025, opened 2019-10-24
RDM-6194-added print api and case activity api services

akdasari:

```
[ ] Yes
[x ] No
```

### DIFF
--- a/k8s/demo/common/ccd/latest-ccd-chart.yaml
+++ b/k8s/demo/common/ccd/latest-ccd-chart.yaml
@@ -34,13 +34,12 @@ spec:
       redirectUri: https://ccd-demo-case-management-web.demo.platform.hmcts.net/oauth2redirect
       s2sUrl: http://ccd-demo-s2s
 
-    # Enable support services tags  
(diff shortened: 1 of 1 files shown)


k8s/demo/common/ccd/latest-ccd-chart.yaml
```diff
@@ -33,6 +33,16 @@ spec:
...
       redirectUri: https://ccd-demo-case-management-web.demo.platform.hmcts.net/oauth2redirect
       s2sUrl: http://ccd-demo-s2s
+
+    # Enable support services tags  
+    ccd:
+      printService:
+        enabled: true
+      activityApi:
+        enabled: true
+      paymentApi:
```
mario-paniccia: paymentApi is not needed, it's already enabled on the product chart.
Also, this whole block should be only on the product chart. Please add a todo to remove it in the future and enable `activityApi` and `printService` permanently on the product chart

akdasari: fixed comments

PR #1053, opened 2019-10-28
Added Prometheus to Mgmt Sandbox

gbstan92:

[AB#504](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/504) 

### Change description ###

Added Prometheus Operator to the new Mgmt Sandbox cluster.

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/mgmt-sandbox/common/monitoring/prom-operator.yaml
+++ b/k8s/mgmt-sandbox/common/monitoring/prom-operator.yaml
@@ -11,7 +11,7 @@ spec:
   releaseName: prom-operator
   forceUpgrade: true
   chart:
-    repository: https://kubernetes-charts.storage.googleapis.com/ 
+    repository: https://kubernetes-charts.storage.googleapis.com/
     name: prometheus-operator
     Version: 6.21.0
   values:
--- a/k8s/mgmt-sandbox/common/monitoring/prom-operator.yaml
+++ b/k8s/mgmt-sandbox/common/monitoring/prom-operator.yaml
@@ -13,7 +13,7 @@ spec:
   chart:
     repository: https://kubernetes-charts.storage.googleapis.com/
     name: prometheus-operator
-    Version: 6.21.0
+    version: 6.21.0
   values:
     defaultRules:
       create: true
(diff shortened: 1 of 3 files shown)


k8s/mgmt-sandbox/common/monitoring/prom-operator.yaml
```diff
@@ -0,0 +1,41 @@
...
+  name: prom-operator
+  namespace: monitoring
+  annotations:
+    flux.weave.works/automated: "false"
+    flux.weave.works/ignore: "false"
+spec:
+  releaseName: prom-operator
+  forceUpgrade: true
+  chart:
+    git: https://github.com/hmcts/charts
```
timja: why is the chart coming from here and not from stable upstream?

timja: i.e. 
https://github.com/hmcts/cnp-flux-config/blob/2f5e2878ee7e5715a3708debbd293f5cdcf28df4/k8s/preview/sealed-secrets/sealed-secrets.yaml#L11

`repository: https://kubernetes-charts.storage.googleapis.com/`

timja: https://github.com/helm/charts/blob/master/stable/prometheus-operator/Chart.yaml

naikajah: I have been told that we are using the hmct fork of the stable chart repository since we have some customizations. So in order to make it consistent across the program, I think we should use the hmcts fork rather than the official.

timja: no that's not right, forking is the last resort, and should only be done in where necessary. we don't want to have to maintain it and especially pointing at master is a very bad idea, next time we updated it this would've been automatically updated to the latest even if there were breaking changes.

naikajah: I definitely agree with that to use the official repository and not using the fork. The usage of master branch depends on the repository you are cloning and what you want to do with that repository. But in general yes we should be using the released tags. 
Also, we should not be practicing inconsistencies as that leads to issues later. If we are using the official repo then use the same everywhere. I only relayed what I was told earlier and hence used the hmcts fork in my POC as well.

timja: the only one I am aware of that is legitimately is forked is the logstash chart, there's an open PR to upstream for our change, but it's been months and no response.

PR #1058, opened 2019-10-28
Configured Prometheus with Ingress on Sandbox

gbstan92:

[AB#504](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/504)

### Change description ###

Added Prometheus Ingress settings on Sandbox. Had to move the config file from common to each cluster individually to avoid using kustomization as it's not required for the purposes of this demo. 

```
[ ] Yes
[x] No
```


k8s/sandbox/cluster-01/monitoring/prom-operator.yaml
```diff
@@ -0,0 +1,49 @@
...
+    kubelet:
+      serviceMonitor:
+        https: false
+
+    grafana:
+      enabled: false
+
+    ingress: 
+      enabled: true
+      hosts: prometheus-01..service.core-compute-sandbox.internal
```
timja:
```suggestion
      hosts: prometheus-01.service.core-compute-sandbox.internal
```

PR #1098, opened 2019-10-30
Removed alias related lines

yogesh-hullatti:
https://tools.hmcts.net/jira/browse/EM-2258

### DIFF
--- a/k8s/demo/common/dg-docassembly/dg-docassembly-demo.yaml
+++ b/k8s/demo/common/dg-docassembly/dg-docassembly-demo.yaml
@@ -17,11 +17,11 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: dg-docassembly
-    version: 0.0.4
+    version: 0.0.5
   values:
     ### EM SERVICES ###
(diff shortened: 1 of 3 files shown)


k8s/demo/common/dg-docassembly/dg-docassembly-demo.yaml
```diff
@@ -17,11 +17,11 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: dg-docassembly
-    version: 0.0.4
+    version: 0.0.5
```
linusnorton: why is this version so far behind?

yogesh-hullatti: I am not sure if this version number should match the one in the microservice chart. I believe this product chart is not updated as frequently as the other.

yogesh-hullatti: > why is this version so far behind?

I am not sure if this version number should match the one in the microservice chart. I believe this product chart is not updated as frequently as the other.

PR #1106, opened 2019-10-31
Create a standalone processor product chart

lgonczar:
### Change description ###

Create a product chart with processor-only to be used by Exela to get SAS token

Depends on https://github.com/hmcts/chart-bulk-scan/pull/5

```
[ ] Yes
[x] No
```


k8s/demo/common/bsp/processor-with-ingress.yaml
```diff
@@ -0,0 +1,37 @@
...
+    enable: true
+    force: true
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: bulk-scan
+    version: 0.1.3
+  values:
+    bulk-scan-processor:
+      java:
+        ingressHost: ""
```
doncem: so ingress is not needed for this one? filename says "with ingress"

mdayican: we will use azure app management?

doncem: then there's no need of having separate processor. it might be only me. perhaps it's easier/safer to assign single processor 🤷‍♂ 

but yea. i think then it's ok

mdayican: change the file name

PR #1118, opened 2019-11-01
Add bulk-scan orchestrator prod 

mdayican:

https://tools.hmcts.net/jira/browse/BPS-867

### Change description ###

add bulks scan orchestrator to prod
 

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/prod/common/bsp/bulk-scan-orchestrator.yaml
+++ b/k8s/prod/common/bsp/bulk-scan-orchestrator.yaml
@@ -14,16 +14,14 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: bulk-scan-orchestrator
-    version: 0.2.0
+    version: 0.2.1
   values:
     java:
       replicas: 2
       useInterpodAntiAffinity: true
       image: hmctspublic.azurecr.io/bulk-scan/orchestrator:latest
       environment:
-        CCD_LOG_LEVEL: none
         IDAM_API_URL: "https://idam-api.platform.hmcts.net"
-        IDAM_CLIENT_REDIRECT_URI: "https://rpe-bulk-scan-processor-prod.service.core-compute-prod.internal/oauth2/callback"
     global:
       environment: prod
       tenantId: "531ff96d-0ae9-462a-8d2d-bec7c0b42082"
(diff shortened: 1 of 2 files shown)


k8s/prod/common/bsp/bulk-scan-orchestrator.yaml
```diff
@@ -0,0 +1,28 @@
...
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: bulk-scan-orchestrator
+    version: 0.2.0
+  values:
+    java:
+      replicas: 2
+      useInterpodAntiAffinity: true
+      image: hmctspublic.azurecr.io/bulk-scan/orchestrator:latest
+      environment:
+        CCD_LOG_LEVEL: none
```
timja: this correct? not INFO?

mdayican: this is sth custom, none correct.

doncem: This was introduced only for testing reasons. We should remove this setting and all custom integration around it

k8s/prod/common/bsp/bulk-scan-orchestrator.yaml
```diff
@@ -0,0 +1,30 @@
...
+        IDAM_API_URL: "https://idam-api.platform.hmcts.net"
+        IDAM_CLIENT_REDIRECT_URI: "https://rpe-bulk-scan-processor-prod.service.core-compute-prod.internal/oauth2/callback"
```
doncem: I wonder why this url needed. The uri is not valid - host doesn't exist and even if it did - endpoint doesn't. My guess it is present because IDAM requires it but no use 🤷‍♂. Just wondering here

k8s/prod/common/bsp/bulk-scan-orchestrator.yaml
mdayican: default is full , you mean default should be none ?

doncem: I mean this config should go away completely: from here and orchestrator service itself. Doesn't block this release, but definitely not needed anymore

mdayican: removed from application and here

k8s/prod/common/bsp/bulk-scan-orchestrator.yaml
mdayican: not needed, removed

PR #1124, opened 2019-11-04
enable dm blobstorage in ccd demo

smathangi:


k8s/demo/common/ccd/latest-ccd-chart.yaml
```diff
@@ -120,6 +120,8 @@ spec:
           MAX_FILE_SIZE: '100MB'
           ENABLE_AZURE_STORAGE_CONTAINER: 'true'
           POSTGRES_STORAGE_ENABLED: 'false'
+      blobstorage:
```
adusumillipraveen: Conditions are root level , not sub chart level. But not sure if enabling at root level might enable it in any other apps. This has been worked out in bulk scan using specific conditions: https://github.com/hmcts/bulk-scan-processor/blob/master/charts/bulk-scan-processor/requirements.yaml#L8

PR #1127, opened 2019-11-04
RDM-6193-rm-Persis

nathan-clark:

```
[ ] Yes
[x] No
```


k8s/demo/common/ccd/latest-ccd-chart.yaml
```diff
@@ -109,10 +109,7 @@ spec:
           MANAGEMENT_WEB_URL: www-ccd-aks.demo.platform.hmcts.net
 
     #Dependencies
-#    TODO enable persistence on chart-ccd and remove it here
```
mario-paniccia: before we do this don't we have to release a new version of the ccd-chart with persistence enabled and use that new version here?

nathan-clark: ?? will check with Anand

PR #1160, opened 2019-11-07
added dm-store ingress config

smathangi:


k8s/demo/common/ccd/latest-ccd-chart.yaml
```diff
@@ -120,6 +123,8 @@ spec:
 #    TODO move dm-store on chart-ccd and remove it here
     dm-store:
       java:
+        ingressClass: traefik-no-proxy
+        ingressHost: dm-store-ccd-aks.demo.platform.hmcts.net
```
mario-paniccia: can we use the global here?

smathangi: done

PR #1179, opened 2019-11-08
Add tests sealed secrets to prod

luigibk:
### Change description ###

```
[ ] Yes
[x] No
```


k8s/prod/common/bar/tests-values.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: bitnami.com/v1alpha1
```
timja: prob want to update codeowners for this file pattern

luigibk: Yeah, even if I suspect it will have to be us anyway :-)

timja: i mean add us

CODEOWNERS
```diff
@@ -49,16 +52,21 @@ k8s/aat/common/ccd/definition-store-api.yaml @hmcts/cdm
 k8s/aat/common/ccd/user-profile-api.yaml @hmcts/cdm
 k8s/perftest/common/ccd/user-profile-api.yaml @hmcts/cdm
 k8s/perftest/common/ccd/definition-store-api.yaml @hmcts/cdm
+k8s/aat/common/ccd/tests-values.yaml @hmcts/cdm
```
timja: can't we just set **/tests-values.yaml to platform-engineering?

PR #1254, opened 2019-11-12
Make bulk scan point to the right blob storage instance in demo

lgonczar:
### Change description ###

Make bulk scan point to the right (static) blob storage instance in demo. Currently, it creates one for itself.

```
[ ] Yes
[x] No
```


k8s/demo/common/bsp/bsp-demo.yaml
```diff
@@ -17,12 +17,25 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: bulk-scan
-    version: 0.1.2
+    version: 0.1.5
```
mdayican: now version: 0.1.6 . available and does not create blob storage.

lgonczar:
Reduced this PR to necessary changes

PR #1265, opened 2019-11-13
enable service bus and use one blob storage

mdayican:
### Change description ###

enable service bus and use one blob storage

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/demo/common/bsp/bsp-sample-app.yaml
+++ b/k8s/demo/common/bsp/bsp-sample-app.yaml
@@ -60,6 +60,16 @@ spec:
     bulk-scan-processor:
       java:
         image: hmctspublic.azurecr.io/bulk-scan/processor:prod-732e2ec3
+        secrets:
+          STORAGE_ACCOUNT_NAME:
+            secretRef: bsp-storage-account-creds
+            key: storageAccountName
+          STORAGE_KEY:
+            secretRef: bsp-storage-account-creds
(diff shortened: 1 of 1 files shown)


k8s/demo/common/bsp/bsp-sample-app.yaml
```diff
@@ -60,6 +60,16 @@ spec:
     bulk-scan-processor:
       java:
         image: hmctspublic.azurecr.io/bulk-scan/processor:prod-732e2ec3
+        secrets:
```
lgonczar: Why do we want these secrets here and not in bsp-demo?

mdayican: do you mean why no in product chart?

doncem: yes. in order to create a sample app product chart one needs to execute several steps and allocate resources we don't actually need - it's a sample app. we should not care about it. just a proof e2e actually works giving the configuration

mdayican: still we need to manage dependencies and configure them. 
or we can use bulk-scan product chart and add sample app.

lgonczar: 👍

mdayican:
I merging this, we will discuss tomorrow  .

PR #1271, opened 2019-11-13
Add rd demo

akiohmcts:


k8s/demo/common/rd/rd-demo.yaml
```diff
@@ -0,0 +1,27 @@
...
+    name: rd
+    version: 0.1.0
+  values:
+    global:
+      idamApiUrl: "https://idam-api.demo.platform.hmcts.net"
+      idamWebUrl: "https://idam-web-public.demo.platform.hmcts.net"
+    rpe-service-auth-provider:
+      java:
+        ingressHost: refdata-s2s-payment-api.demo.platform.hmcts.net
+        ingressClass: traefik-no-proxy
```
timja: what is this file:
sealed-secrets/s2s-url.yaml

also you don't need s2s to be exposed publically? remove ingress class and ingress host

akiohmcts: sealed-secrets was created using the shell script you generated for fees-pay to store Vault secrets in an encrypted format... I just forgot to remove it. It seems like we won't need to port the vault secrets from rd-demo vault as this is a self-contained environment with all the dependencies being deployed there - but correct me if I'm wrong.

PR #1272, opened 2019-11-13
set service bus connection string

mdayican:
### Change description ###

set service bus connection string

```
[ ] Yes
[x] No
```


k8s/demo/common/bsp/bsp-sample-app.yaml
```diff
@@ -67,12 +67,19 @@ spec:
           STORAGE_KEY:
             secretRef: bsp-storage-account-creds
             key: accessKey
+          SB_CONN_STRING:
+            secretRef: servicebus-secret-namespace-{{ .Release.Name }}-sb
```
lgonczar: I wonder if we could make something global out this at some point.

PR #1423, opened 2019-11-25
RDM-6581-fixed service enablement as per latest chart

akdasari:
### Change description ###
Enabled all services on Demo CCD

```
[ ] Yes
[ x] No
```

### DIFF
--- a/k8s/demo/common/ccd/latest-ccd-chart.yaml
+++ b/k8s/demo/common/ccd/latest-ccd-chart.yaml
@@ -82,6 +82,11 @@ spec:
       java:
         memoryRequests: "1G"
         memoryLimits: "2G"
+        secrets:
+          STORAGE_ACCOUNT_NAME:
+            disabled: false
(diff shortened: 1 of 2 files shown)


k8s/demo/common/ccd/latest-ccd-chart.yaml
```diff
@@ -21,6 +21,27 @@ spec:
...
+        enabled: true
+      dmStore:
+        enabled: true
+      managementWeb:
+        enabled: true
+      apiGatewayWeb:
+        enabled: true
+      activityApi:
+        enabled: true
+      blobstorage:
```
mario-paniccia: when you enable `blobstorage` there is a property of def store you have to set to true so that we can see the list of imports in Admin Web: AZURE_STORAGE_DEFINITION_UPLOAD_ENABLED.
Please ask Ola to verify that this worked and we can see the list of imports on Admin Web

PR #1429, opened 2019-11-25
ignore bsp-demo as bsp-sample-app is same

mdayican:
### Change description ###

stop deploying bsp-demo as bsp-sample-app is same

```
[ ] Yes
[x] No
```


lgonczar:
I wonder if we should just remove the file instead.

mdayican:
> I wonder if we should just remove the file instead.

let's first test?

PR #1485, opened 2019-11-28
Setup for sscs callback orchestrator on prod

jackmaloney:


timja:
> Is it possible to get the application insight key moved to a secret? also the tenant? the tenant less of an issue (but it does give attackers a target) however the application instrumentation key can be used by anyone to send data in, potentially leading to a DOS attack on the logs, or a flood to cover/mask a targeted attack.

example for app insights key:
https://github.com/hmcts/sscs-case-loader/blob/master/charts/sscs-case-loader/values.yaml#L26
https://github.com/hmcts/sscs-case-loader/blob/master/src/main/resources/bootstrap.yaml#L23

most apps are getting it from vault but there's a few that are putting it inn directly.

the tenant ID isn't possible in the current setup, it's everywhere

jackmaloney:
Updated @noferrets

PR #1617, opened 2019-12-10
Remove FPL case service from performance test env

geek-soft:

None

### Change description ###

As per title.

```
[ ] Yes
[x] No
```


geek-soft:
They can remain in case we want need to use env again. The whole point is to remove need for maintenance of that rarely used env.

PR #1631, opened 2019-12-11
Removing access to BSP endpoints in PROD

jakub-kozlowski:

https://tools.hmcts.net/jira/browse/BSP-140

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/k8s/prod/common/divorce/div-cos.yaml
+++ b/k8s/prod/common/divorce/div-cos.yaml
@@ -41,6 +41,8 @@ spec:
         DOCUMENTATION_SWAGGER_ENABLED: "false"
         FEATURE_RESP_SOLICITOR_DETAILS: "true"
         SCHEDULER_SEND_UPDATED_CASES_TO_ROBOTICS_CRON: "0 0 4 ? * * *"
+        BSP_SERVICES_ALLOWED_TO_UPDATE:
+        BSP_SERVICES_ALLOWED_TO_VALIDATE:
       aadIdentityName: divorce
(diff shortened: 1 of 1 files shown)


k8s/prod/common/divorce/div-cos.yaml
```diff
@@ -41,6 +41,8 @@ spec:
         DOCUMENTATION_SWAGGER_ENABLED: "false"
         FEATURE_RESP_SOLICITOR_DETAILS: "true"
         SCHEDULER_SEND_UPDATED_CASES_TO_ROBOTICS_CRON: "0 0 4 ? * * *"
+        BSP_SERVICES_ALLOWED_TO_UPDATE:
```
yarpo: Hm, maybe 

```
BSP_SERVICES_ALLOWED_TO_UPDATE: ""
```
?

jakub-kozlowski: that would be empty string, which is not the same as empty value - we won't know before we test :)

jakub-kozlowski: I'd try with empty value and only change it to empty string if it doesn't work

yarpo: yeah, but we use it this way:

`${BSP_SERVICES_ALLOWED_TO_UPDATE:default_value}`

if BSP_SERVICES_ALLOWED_TO_UPDATE is not set -> default_value. I guess.

PR #1639, opened 2019-12-12
fix for s2s poddisruptionbudget issue

naikajah:


k8s/demo/common/ccd/ccd-full-with-pay.yaml
```diff
@@ -65,6 +65,8 @@ spec:
           enabled: false
         ccpay-bulkscanning-api:
           enabled: false
+        ccd:
```
smathangi: ccd is enabled by default from ccpay chart?

naikajah: yes

PR #1728, opened 2020-01-07
Update the vault name

yogesh-hullatti:
Update the vault name


yogesh-hullatti:
Not required, as we will be renaming the vault name

PR #1756, opened 2020-01-09
fix: updated rd-judicial yaml for aat

abhijitdiwan:
NA

### Change description ###
fix: updated rd-judicial yaml for aat

No


k8s/aat/common/rd/judicial-data-load.yaml
```diff
@@ -19,7 +19,7 @@ spec:
     java:
       replicas: 2
       useInterpodAntiAffinity: true
-      image: hmctspublic.azurecr.io/rd/judicial-data-load:prod-dd3f7021
+      image: hmctspublic.azurecr.io/rd/judicial-data-load:latest
```
naikajah: Could you please check if this is the correct syntax? We have asked all the teams not to use **latest**

PR #1835, opened 2020-01-16
Moving to helm 3 on helm operator

adusumillipraveen:

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/admin/helm-release-patch.yaml
+++ b/k8s/admin/helm-release-patch.yaml
@@ -1,0 +1,6 @@
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
+  name: test
+spec:
+  helmVersion: v3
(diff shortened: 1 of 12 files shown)


k8s/admin/helm-release-patch.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
+  name: test
```
adusumillipraveen: This dummy and ignored by kustomize

k8s/sandbox/cluster-00-overlay/kustomization.yaml
```diff
@@ -30,3 +30,8 @@ patchesStrategicMerge:
 #flux-web patch
 - ../../admin/flux-web/patches/sandbox/cluster-00/flux-web.yaml
 
+
+patches:
```
adusumillipraveen: This would patch all helmreleases with version: v3

PR #1864, opened 2020-01-20
add blob router to prod

mdayican:

https://tools.hmcts.net/jira/browse/BPS-797

### Change description ###

add blob router service flux configuration for prod.

```
[ ] Yes
[x] No
```


nitinprabhu:
> probably need to update CODEOWNERS too:
> 
> > k8s/aat/common/reform-scan/blob-router.yaml @hmcts/bulk-scan

good point we always miss that

PR #1917, opened 2020-01-24
add flux config for xui

ernestman28:

### Change description ###

```
[ ] Yes
[ ] No
```


CODEOWNERS
```diff
@@ -199,6 +199,17 @@ k8s/ithc/common/xui/webapp.yaml @hmcts/xui
 k8s/ithc/common/xui/manage-organisations.yaml @hmcts/xui
 k8s/ithc/common/xui/approve-org.yaml @hmcts/xui
 k8s/ithc/common/xui/share-case.yaml @hmcts/xui
+k8s/perftest/common/xui/xui-tc.yaml @hmcts/xui
+k8s/perftest/common/xui/webapp.yaml @hmcts/xui
+k8s/perftest/common/xui/manage-organisations.yaml @hmcts/xui
+k8s/perftest/common/xui/approve-org.yaml @hmcts/xui
+k8s/perftest/common/xui/share-case.yaml @hmcts/xui
+k8s/demo/common/xui/xui-tc.yaml @hmcts/xui
```
adusumillipraveen: Demo doesn't need code owners as of now.

k8s/demo/common/xui/xui-tc.yaml
```diff
@@ -0,0 +1,27 @@
+---
+apiVersion: helm.fluxcd.io/v1
```
adusumillipraveen: Do u know demo doesn't work same as other environments with current setup?

CODEOWNERS
ernestman28: okay I will remove this

k8s/demo/common/xui/xui-tc.yaml
ernestman28: I only know we can use Product chart or Microservice chart in demo, but didnt know the actual setup will be different in flux config. Can you point me in the right direction to setup demo env for using microservice chart? thx

adusumillipraveen: I mean environment itself is setup differently. For example, it doesn't have connectivity to existing postgres databases. so, ur config won't wait.

ernestman28: so what we need to do if we would like to connect to a PaaS postgre instance? i.e  xui-terms-and-conditions-demo . Or the demo environment doesnt support connect to any PaaS DB?

adusumillipraveen: It doesn't support connecting to any Paas DB.

PR #1928, opened 2020-01-27
RDM-7389 aat test-stubs-service.yaml

nathan-clark:
Will bump chart when available.

```
[ ] Yes
[x] No
```


k8s/aat/common/ccd/test-stubs-service.yaml
```diff
@@ -0,0 +1,26 @@
...
+  releaseName: ccd-test-stubs-service
+  rollback:
+    enable: true
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: ccd-test-stubs-service
+    version: 1.0.2
+  values:
+    java:
+      replicas: 2
```
smathangi: i guess, one replica is fine for this service. So total of two instances (both clusters)

PR #1973, opened 2020-01-30
Bump payment processor to 0.1.8

rkondratowicz:
https://github.com/hmcts/bulk-scan-payment-processor/pull/171/files


lgonczar:
> fix codeowners in here : https://github.com/hmcts/cnp-flux-config/blob/master/CODEOWNERS#L182

@adusumillipraveen Can we not make this PR dependent on that change? There's more projects in Bulk Scan that need to have Platform Engineering removed

PR #1987, opened 2020-02-03
Enable processing crime container in prod (blob-router-service)

lgonczar:
### Change description ###

Enable processing crime container in prod (blob-router-service)

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/prod/common/reform-scan/blob-router.yaml
+++ b/k8s/prod/common/reform-scan/blob-router.yaml
@@ -21,6 +21,7 @@ spec:
       useInterpodAntiAffinity: true
       image: hmctspublic.azurecr.io/reform-scan/blob-router:prod-d9a7fbc1
       environment:
+        CRIME_ENABLED: true
         HANDLE_REJECTED_FILES_CRON: "0 0 7 * * *"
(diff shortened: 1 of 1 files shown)


k8s/prod/common/reform-scan/blob-router.yaml
```diff
@@ -21,6 +21,7 @@ spec:
       useInterpodAntiAffinity: true
       image: hmctspublic.azurecr.io/reform-scan/blob-router:prod-d9a7fbc1
       environment:
+        CRIME_ENABLED: true
```
lgonczar: Alternatively, we could just have crime enabled by default. Thoughts?

nitinprabhu: Crime processing won’t work on prod at the moment so better not to enable it for now ?

We could enable it by default in values yaml and disable for prod in flux ?

doncem: if not going to work on prod why are we enabling then here?

lgonczar: Why is it not going to work in prod?

PR #2082, opened 2020-02-11
Enable MYA feature flags on prod

jackmaloney:


k8s/perftest/common/sscs/sscs-tya-notif.yaml
```diff
@@ -14,7 +14,7 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: sscs-tya-notif
```
srochani: Why tya app? I think it should be core-front end

jackmaloney: this is notification service

PR #2098, opened 2020-02-12
Flux helm operator rollback retry rpe aat

luigibk:

[[AB#781](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/781)]

### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/aat/common/rpe/send-letter-service.yaml
+++ b/k8s/aat/common/rpe/send-letter-service.yaml
@@ -11,6 +11,8 @@ spec:
   releaseName: rpe-send-letter-service
   rollback:
     enable: true
+    retry: true
+    maxRetries: 5
   chart:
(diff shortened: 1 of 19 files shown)


k8s/aat/common/rpe/send-letter-service.yaml
```diff
@@ -11,6 +11,8 @@ spec:
   releaseName: rpe-send-letter-service
   rollback:
     enable: true
+    retry: true
+    maxRetries: 5
```
adusumillipraveen: This should be default ?
```suggestion
```

luigibk: it's meant for teams to know that the option exists ... but, on second thought, do we want them to know?

PR #2119, opened 2020-02-14
Deploying COS to demo

jvictorcf:
### Change description ###
Deploying COS to Demo

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/k8s/demo/common/divorce/div-cos.yaml
+++ b/k8s/demo/common/divorce/div-cos.yaml
@@ -27,7 +27,7 @@ spec:
       readinessTimeout: 5
       readinessPeriod: 15
       useInterpodAntiAffinity: true
-      applicationInsightsInstrumentKey: "63c30ef9-a27b-4d38-8afc-4c180fb6ac33"
+      applicationInsightsInstrumentKey: "686ed071-9c72-4dc3-9edb-191edceb6018"
       image: hmctspublic.azurecr.io/div/cos:prod-8f87f4d9
       environment:
         IDAM_API_URL: "https://idam-api.demo.platform.hmcts.net"
--- a/k8s/demo/common/divorce/div-cos.yaml
+++ b/k8s/demo/common/divorce/div-cos.yaml
@@ -1,0 +1,51 @@
+---
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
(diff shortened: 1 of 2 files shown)


k8s/demo/common/divorce/div-cos.yaml
```diff
@@ -0,0 +1,51 @@
...
+    enable: true
+    retry: true
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: div-cos
+    version: 1.3.2
+  values:
+    java:
+      replicas: 2
+      memoryRequests: '768Mi'
```
timja: you shouldn't need these memory limited and cpu requests

jvictorcf: Why? Are these the default values?
Should I remove it only from the demo file or from all of them?

k8s/demo/common/divorce/div-cos.yaml
```diff
@@ -0,0 +1,51 @@
...
+      cpuRequests: '1000m'
+      memoryLimits: '3072Mi'
+      cpuLimits: '2500m'
+      readinessDelay: 45
+      readinessTimeout: 5
+      readinessPeriod: 15
+      useInterpodAntiAffinity: true
+      applicationInsightsInstrumentKey: "63c30ef9-a27b-4d38-8afc-4c180fb6ac33"
```
timja: is this the demo key?

k8s/demo/common/divorce/div-cos.yaml
timja: tbh, not sure the history, 95% of apps don't need it, it was probably copied from the original, but might be too much work to find out

k8s/demo/common/divorce/div-cos.yaml
jvictorcf: Good question. Let me find out.

jvictorcf: It wasn't. Good catch.
Changed it now.

k8s/demo/common/divorce/div-cos.yaml
jvictorcf: fair enough. I'll leave it, in that case.

PR #2195, opened 2020-02-20
IA chart bump for documents and restart AAT

makowalski:
N/A

### Change description ###
IA chart bump for documents and restart AAT

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/aat/common/ia/case-api.yaml
+++ b/k8s/aat/common/ia/case-api.yaml
@@ -22,6 +22,8 @@ spec:
       useInterpodAntiAffinity: true
       image: hmctspublic.azurecr.io/ia/case-api:prod-dbff1f2f
       ingressHost: ia-case-api-aat.service.core-compute-aat.internal
+      environment:
+        RESTART_ME: "Update due to vault change"
     global:
       environment: aat
       tenantId: "531ff96d-0ae9-462a-8d2d-bec7c0b42082"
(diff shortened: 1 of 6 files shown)


k8s/aat/common/ia/case-api.yaml
```diff
@@ -22,6 +22,8 @@ spec:
       useInterpodAntiAffinity: true
       image: hmctspublic.azurecr.io/ia/case-api:prod-dbff1f2f
```
jkumar-moj: Don't we need version bump here?

k8s/aat/common/ia/case-notifications-api.yaml
```diff
@@ -24,7 +24,7 @@ spec:
       image: hmctspublic.azurecr.io/ia/case-notifications-api:prod-d3199230
       ingressHost: ia-case-notifications-api-aat.service.core-compute-aat.internal
```
jkumar-moj: Don't we need version bump here?

k8s/aat/common/ia/case-api.yaml
makowalski: nope, version is bumped only for documents api, case-api needs only restart

existing version 0.0.5 matches https://github.com/hmcts/ia-case-api/blob/master/charts/ia-case-api/Chart.yaml#L4
version in application

jkumar-moj: Thanks Marcin..

PR #2216, opened 2020-02-21
AKS Migrate to AAT - CCD Case Print Service

naikajah:


k8s/aat/common/ccd/case-print-service.yaml
```diff
@@ -0,0 +1,29 @@
...
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: ccd-case-print-service
+    version: 1.0.4
+  values:
+    nodejs:
+      replicas: 2
+      useInterpodAntiAffinity: true
+      image: hmctspublic.azurecr.io/ccd/case-print-service:prod-200e9ada
+      environment:
+        IDAM_AUTHENTICATION_WEB_URL: https://idam.preprod.ccidam.reform.hmcts.net
```
timja: this is wrong, it should be, https://idam-web-public.aat.platform.hmcts.net, but won't it be defaulted in the chart?

naikajah: yeah, I thought this to be the same but the existing value which was there was different (the one I have put there). So I was not sure if I should remove that value or not.

naikajah: have updated the file and will check with IDAM team.

PR #2223, opened 2020-02-21
Configure blob-router service for perftest

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-1031

### Change description ###

Configure blob-router service for perftest

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/perftest/common/reform-scan/blob-router.yaml
+++ b/k8s/perftest/common/reform-scan/blob-router.yaml
@@ -1,0 +1,32 @@
+---
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
+  name: reform-scan-blob-router
+  namespace: reform-scan
+  annotations:
+    flux.weave.works/automated: "true"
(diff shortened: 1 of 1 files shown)


k8s/perftest/common/reform-scan/blob-router.yaml
```diff
@@ -0,0 +1,32 @@
+---
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
+  name: reform-scan-blob-router
+  namespace: reform-scan
+  annotations:
+    flux.weave.works/automated: "true"
+    flux.weave.works/tag.java: glob:perftest-*
```
nitinprabhu: This should be prod and not perftest

nitinprabhu:
```suggestion
    flux.weave.works/tag.java: glob:prod-*
```

lgonczar: Got carried away. Well spotted. Done

PR #2247, opened 2020-02-24
Bump blob router

rkondratowicz:
https://github.com/hmcts/blob-router-service/pull/291


rkondratowicz:
bumped perftest as well

PR #2258, opened 2020-02-25
add prod config

naikajah:


k8s/prod/common/ccd/case-print-service.yaml
```diff
@@ -0,0 +1,28 @@
...
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: ccd-case-print-service
+    version: 1.0.4
+  values:
+    nodejs:
+      replicas: 2
+      useInterpodAntiAffinity: true
+      image: hmctspublic.azurecr.io/ccd/case-print-service:prod-689abd7a
+      ingressHost: return-case-doc.ccd.platform.hmcts.net
```
adusumillipraveen: need these? https://github.com/hmcts/ccd-case-print-service/blob/master/charts/ccd-case-print-service/values.yaml#L24-L26

PR #2308, opened 2020-02-28
Disable proxy usage for perftest in blob-router

lgonczar:
### Change description ###

Disable proxy usage for perftest in blob-router

```
[ ] Yes
[x] No
```


k8s/aat/common/reform-scan/blob-router.yaml
```diff
@@ -15,7 +15,7 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: reform-scan-blob-router
-    version: 0.0.31
+    version: 0.0.32
```
aliveni: There is already a PR raised for this version https://github.com/hmcts/cnp-flux-config/pull/2311

PR #2413, opened 2020-03-10
Email reports in AAT

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1026

### Change description ###
This is to test daily email reports in AAT. 
report-recipients value in reform-scan-aat secrets has been set to HMCTS support group
Report is supposed to be sent every 5 minutes.
This PR will be reverted after testing.

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/aat/common/reform-scan/blob-router.yaml
+++ b/k8s/aat/common/reform-scan/blob-router.yaml
@@ -24,6 +24,9 @@ spec:
       environment:
         STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 0
         RESTART_ME: "To recreate pods"
+        SEND_DAILY_REPORT_ENABLED: "true"
+        SMTP_HOST: smtp.office365.com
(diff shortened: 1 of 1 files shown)


k8s/aat/common/reform-scan/blob-router.yaml
```diff
@@ -24,6 +24,9 @@ spec:
       environment:
         STORAGE_BLOB_PROCESSING_DELAY_IN_MINUTES: 0
         RESTART_ME: "To recreate pods"
+        SEND_DAILY_REPORT_ENABLED: "true"
+        SMTP_HOST: smtp.office365.com
+        SEND_DAILY_REPORT_CRON: "* */5 * ? * MON-FRI"
```
nitinprabhu: for testing we can keep it an hour otherwise you will be bombarded with loads of email

apereverzin: @nitinprabhu  Done

PR #2489, opened 2020-03-13
update ithc data store to ithc branch

mario-paniccia:


mario-paniccia:
not needed. We should use the PR image

PR #2664, opened 2020-03-25
Added flux config for FPLA demo

mounikahmcts:
https://tools.hmcts.net/jira/browse/FPLA-597
Added flux config for FPLA demo

### DIFF
--- a/k8s/demo/common/family-public-law/fpl-case-service.yaml
+++ b/k8s/demo/common/family-public-law/fpl-case-service.yaml
@@ -1,0 +1,35 @@
+---
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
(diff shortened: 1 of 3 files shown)


k8s/demo/common/family-public-law/fpl-case-service.yaml
```diff
@@ -0,0 +1,35 @@
...
+    flux.weave.works/tag.java: glob:prod-*
+spec:
+  releaseName: fpl-case-service
+  rollback:
+    enable: true
+    retry: true
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: fpl-case-service
+    version: 1.8.0
```
akomisarek: Latest version is noow 1.9.0 as Sam did some update, please update :)

PR #2801, opened 2020-04-02
Remove app insights key as it is already made dynamic in tf file

mounikahmcts:
Reference:
https://github.com/hmcts/fpl-ccd-configuration/pull/990

### DIFF
--- a/k8s/ithc/common/family-public-law/fpl-case-service.yaml
+++ b/k8s/ithc/common/family-public-law/fpl-case-service.yaml
@@ -14,7 +14,7 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: fpl-case-service
-    version: 1.8.0
+    version: 1.10.4
   values:
     java:
       replicas: 2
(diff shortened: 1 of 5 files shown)


Gogetter:
@mounikahmcts do we need to add this to key vault and reference in `values.yaml`?

mounikahmcts:
> @mounikahmcts do we need to add this to key vault and reference in `values.yaml`?

@Gogetter  - I have already done that as part of https://github.com/hmcts/fpl-ccd-configuration/pull/990/files

k8s/ithc/common/family-public-law/fpl-case-service.yaml
```diff
@@ -14,15 +14,14 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: fpl-case-service
-    version: 1.8.0
+    version: 1.10.4
```
Gogetter, line 17: Looks like we are currently on [1.10.5](https://github.com/hmcts/fpl-ccd-configuration/blob/master/charts/fpl-case-service/Chart.yaml)

mounikahmcts: updated it

PR #2841, opened 2020-04-06
Update chart version in bulk-scan-orchestrator prod

lgonczar:
### Change description ###

Update chart version in bulk-scan-orchestrator prod

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/prod/common/bsp/bulk-scan-orchestrator.yaml
+++ b/k8s/prod/common/bsp/bulk-scan-orchestrator.yaml
@@ -14,7 +14,7 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: bulk-scan-orchestrator
-    version: 0.2.8
+    version: 0.2.9
   values:
     java:
       replicas: 2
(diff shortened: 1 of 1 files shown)


k8s/prod/common/bsp/bulk-scan-orchestrator.yaml
```diff
@@ -24,6 +24,7 @@ spec:
         IDAM_API_URL: "https://idam-api.platform.hmcts.net"
         TRANSFORMATION_URL_DIVORCE: ""
         TRANSFORMATION_URL_FINREM: ""
+        UPDATE_URL_PROBATE: ""
```
lgonczar, line 27: Do you think we need this?

mdayican: if we want to override , if we do not want to use in prod we need.

nitinprabhu: Don't think so we need this as it is derived from value yaml

mdayican: we do not want to use this in prod?

lgonczar: [discussion was moved to dev channel]

PR #2859, opened 2020-04-07
PRO-1234: point demo to PR-1042 for BO

sanjayparekh-hmcts:
Point demo to probate-backoffice PR-1042

### DIFF
--- a/k8s/demo/common/probate/probate-back-office.yaml
+++ b/k8s/demo/common/probate/probate-back-office.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: probate
   annotations:
     fluxcd.io/automated: "true"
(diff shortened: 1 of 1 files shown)


k8s/demo/common/probate/probate-back-office.yaml
```diff
@@ -20,7 +20,7 @@ spec:
     java:
       replicas: 2
       useInterpodAntiAffinity: true
-      image: "hmctspublic.azurecr.io/probate/back-office:prod-67fd917f"
+      image: "hmctspublic.azurecr.io/probate/back-office:pr-1042-880af01d"
```
timja, line 23: i haven't verified this but if you've checked its fine,
next time you can just update the tag filter and flux will automatically find it for you and handle it.

PR #2869, opened 2020-04-08
Add bulk-scan-processor config for demo

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-1161

### Change description ###

Add bulk-scan-processor config for demo

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/demo/common/bsp/bulk-scan-processor.yaml
+++ b/k8s/demo/common/bsp/bulk-scan-processor.yaml
@@ -15,7 +15,7 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: bulk-scan-processor
-    version: 0.2.19
+    version: 0.2.20
   values:
     java:
       replicas: 2
--- a/k8s/demo/common/bsp/bulk-scan-processor.yaml
+++ b/k8s/demo/common/bsp/bulk-scan-processor.yaml
@@ -1,0 +1,33 @@
+---
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
+  name: bulk-scan-processor
+  namespace: bsp
+  annotations:
+    flux.weave.works/automated: "true"
+    flux.weave.works/tag.java: glob:prod-*
+spec:
+  releaseName: bulk-scan-processor
+  rollback:
+    enable: true
+    retry: true
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: bulk-scan-processor
+    version: 0.2.19
+  values:
+    java:
+      replicas: 1
+      useInterpodAntiAffinity: true
+      cpuRequests: '500m'
+      cpuLimits: '2500m'
+      image: hmctspublic.azurecr.io/bulk-scan/processor:prod-f01fe889
+      environment:
+        STORAGE_BLOB_SELECTED_CONTAINER: ALL
+        #60 days in seconds
+        SAS_TOKEN_VALIDITY: 5184000
+    global:
+      environment: demo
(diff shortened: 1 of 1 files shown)


k8s/demo/common/bsp/bulk-scan-processor.yaml
```diff
@@ -0,0 +1,33 @@
...
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: bulk-scan-processor
+    version: 0.2.19
+  values:
+    java:
+      replicas: 1
+      useInterpodAntiAffinity: true
+      cpuRequests: '500m'
+      cpuLimits: '2500m'
+      image: hmctspublic.azurecr.io/bulk-scan/processor:prod-f01fe889
```
mdayican, line 25: you can use :latest

lgonczar: Does it make any difference if this is going to be updated automatically?

doncem: my guess would be "just in case it will not be tagged something as _prod-..._"

lgonczar: What you you mean, Don?

doncem: i mean in case that tag is not found and flux never deploys anything and maybe never auto-updates. just hypothesis here. i have no problem with what it is now ;)

mdayican: I thought it may first deploy the one stated in the config with even there is a newer version. it should be fine that's why I approved :)

lgonczar: Is this correct?
```
image: hmctspublic.azurecr.io/bulk-scan/processor:latest
```

mdayican: yes, after deployment you will see flux updated with the image tag

lgonczar: Thanks for explaining, guys. Done.

k8s/demo/common/bsp/bulk-scan-processor.yaml
```diff
@@ -0,0 +1,33 @@
...
+  rollback:
+    enable: true
+    retry: true
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: bulk-scan-processor
+    version: 0.2.19
+  values:
+    java:
+      replicas: 1
```
nitinprabhu, line 21: any reason to have just one replica ?

lgonczar: mistake, fixed

lgonczar: Default capacity we've got in processor for ASE is 1 and it's only different in aat and prod. I guess that's where it comes from. Other configs for AKS demo have 2, so it make sense to stick with 2.

k8s/demo/common/bsp/bulk-scan-processor.yaml
```diff
@@ -0,0 +1,33 @@
...
+    flux.weave.works/tag.java: glob:prod-*
+spec:
+  releaseName: bulk-scan-processor
+  rollback:
+    enable: true
+    retry: true
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: bulk-scan-processor
+    version: 0.2.19
```
nitinprabhu, line 18: processor is now on 0.2.20

lgonczar: Updated. I'm assuming this version will come soon.

PR #2896, opened 2020-04-09
Revert "bulk-scan-processor version 0.2.20"

doncem:
breaks flux as still no chart with that version. have to re-reverted later

### DIFF
--- a/k8s/perftest/common/bsp/bulk-scan-processor.yaml
+++ b/k8s/perftest/common/bsp/bulk-scan-processor.yaml
@@ -14,7 +14,7 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: bulk-scan-processor
-    version: 0.2.20
+    version: 0.2.18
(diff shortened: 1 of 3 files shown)


k8s/perftest/common/bsp/bulk-scan-processor.yaml
```diff
@@ -14,7 +14,7 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: bulk-scan-processor
-    version: 0.2.20
+    version: 0.2.18
```
timja, line 17: did you mean for this to go to 18?

nitinprabhu: should be 19 I think

doncem: ooh. didn't look at diff. just clicked revert button on pr landing page ;). lemme push extra commit

PR #3014, opened 2020-04-21
blob-router-service 0.0.42

apereverzin:
### Change description ###
Enabled for prod

```
[ ] Yes
[x] No
```


doncem:
> SMTP_HOST config is missing ?

ah, should've checked the blob router after this pr has changed. yea, missing smtp host

PR #3040, opened 2020-04-22
Moving artifactory to new mgmt clusters

adusumillipraveen:
Sticking to lift and shift currently , it doesn't have a chart.

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/namespaces/artifactory/atrifactory.yaml
+++ b/k8s/namespaces/artifactory/atrifactory.yaml
@@ -1,0 +1,95 @@
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: artifactory-oss
+  namespace: artifactory
+spec:
+  serviceName: "artifactory"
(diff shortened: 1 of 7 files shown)


k8s/namespaces/artifactory/atrifactory.yaml
```diff
@@ -0,0 +1,95 @@
...
+              name: artifactory-data
+  volumeClaimTemplates:
+    - metadata:
+        name: artifactory-data
+      spec:
+        accessModes: [ "ReadWriteOnce" ]
+        storageClassName: "managed-premium"
+        resources:
+          requests:
+            storage: 1000Gi
```
timja, line 73: do we need such a big disk? last time we checked it wasn't using anywhere near that

adusumillipraveen: leaving it for now

PR #3077, opened 2020-04-23
Add admin for FPL namespace to authenticated users - POC

akomisarek:
### Change description ###
As in title. 

```
[ ] Yes
[X] No
```

### DIFF
--- a/k8s/aat/common/family-public-law/rolebinding.yaml
+++ b/k8s/aat/common/family-public-law/rolebinding.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: admin-fpl
+  namespace: family-public-law
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: admin
+subjects:
+  - kind: Group
+    name: system:authenticated
+    apiGroup: rbac.authorization.k8s.io


k8s/aat/common/family-public-law/rolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: admin-fpl
+  namespace: family-public-law
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: admin
+subjects:
+  - kind: Group
+    name: system:authenticated
```
timja, line 12: this shouldn't be authenticated it should be the user group

timja:
Has this been approved anywhere? (I'm just looking at it from a code perspective)

akomisarek:
> Has this been approved anywhere? (I'm just looking at it from a code perspective)

Not yet, I am exploring options and testing if that even works - want to open DevOps ticket with problems and investigations we did. Part of it was checking if dev team is empowered to do changes on that level :), which seems not as it is still blocked.

k8s/aat/common/family-public-law/rolebinding.yaml
akomisarek: Yes, but i guess it needs to be reference by objectId?

akomisarek:
> this doesn't seem to scope it to fpl at all?

I updated the file, not sure if you seen latest version - it binds the role only to FPL namespace.

k8s/aat/common/family-public-law/rolebinding.yaml
akomisarek: Yes I know, do we have way of dynamically extracting it?

akomisarek:
Ideally I would like to enable extra functionality on namespaces where the teams have got a bit more ownership: i.e. exec on the pods. I was thinking about ClusterRole:
`apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: exec-pod
rules:
- apiGroups: [""]
  resources: ["pods/exec"]
  verbs: ["create"]`

And then binding to that role on this level.

timja:
not sure how much benefit exec will give your normally as you're running an image without a shell in most cases, port-forward would be more useful and delete pod / hr.

akomisarek:
Thanks Tim for input, yeah in this instance it would be helpful for threaddump, it seems there is jattach on those images so would be good to use it, but yeah maybe generally port-forward is better, we could connect to JMX even 🤔

timja:
> Thanks Tim for input, yeah in this instance it would be helpful for threaddump, it seems there is jattach on those images so would be good to use it, but yeah maybe generally port-forward is better, we could connect to JMX even 🤔

ah forgot about that, it makes sense to keep exec then

k8s/aat/common/family-public-law/rolebinding.yaml
timja: we look it up in the pipeline and put it in, you would likely need to hardcode it here

PR #3111, opened 2020-04-27
Enable sending daily reports in prod for blob-router

lgonczar:
### Change description ###

Enable sending daily reports in prod for blob-router. Default cron is used: `0 0 18 ? * MON-FRI`

```
[ ] Yes
[x] No
```


lgonczar:
> need smtp host configured
> 
> `SMTP_HOST: smtp.office365.com`

Done. Well spotted.

PR #3152, opened 2020-04-28
Temporarily disable sending notifications on prod

nitinprabhu:


lgonczar:
What if we just disable scan?

nitinprabhu:
> What if we just disable scan?

yeah could do that but we would have to do disable `NO_NEW_ENVELOPES_TASK_ENABLED` otherwise we may lot of errors

PR #3249, opened 2020-05-04
Added ecm yaml files

jacarrete:
Added ecm yaml files


k8s/aat/common/ethos/ecm-consumer.yaml
```diff
@@ -0,0 +1,29 @@
...
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: ecm-consumer
+    version: 0.0.1
+  values:
+    java:
+      replicas: 2
+      useInterpodAntiAffinity: true
+      image: hmctspublic.azurecr.io/ethos/ecm-consumer:aat-eb96e9bc
+      ingressHost: ecm-consumer-aat.service.core-compute-aat.internal
```
timja, line 24: this should be in the chart really and templated with the environment

jacarrete: yeah it is in the chart, https://github.com/hmcts/ecm-consumer/blob/26fb1a5b18f557f81c4602fd2bd94dfcc8dc0402/charts/ecm-consumer/values.yaml#L4, i guess i can take that out from the aat and prod yaml right?

timja: yes from all of them

jacarrete: updated!

CODEOWNERS
```diff
@@ -214,6 +214,7 @@ k8s/perftest/common/am/accessmgmt-api.yaml @hmcts/am-devs
 k8s/**/common/ethos/namespace.yaml @hmcts/ethos
 k8s/**/common/ethos/identity.yaml @hmcts/ethos
 k8s/**/common/ethos/repl-docmosis-service.yaml @hmcts/ethos
+k8s/**/common/ethos/ecm-consumer.yaml @hmcts/ethos
```
timja, line 217: if it's not allowed in prod then this shouldn't be **

```suggestion
k8s/aat/common/ethos/ecm-consumer.yaml @hmcts/ethos
```

jacarrete: I deleted the yaml file from prod. It should not be found anymore.

jacarrete: > if it's not allowed in prod then this shouldn't be **

 deleted the yaml file from prod. It should not be found anymore.

PR #3270, opened 2020-05-05
Bsp and Reform-scan just use demo 00

mdayican:
### Change description ###

 as there is a synchronisation problem between demo clusters all BSP and Refrom-scan components moved to demo-00

```
[ ] Yes
[x] No
```


k8s/demo/cluster-00/bsp/bsp-sample-app.yaml
```diff
@@ -0,0 +1,143 @@
...
+                secretRef: storage-secret-{{ .Release.Name }}
+                key: storageAccountName
+              STORAGE_ACCOUNT_KEY:
+                secretRef: storage-secret-{{ .Release.Name }}
+                key: accessKey
+      ccpay:
+        payment-api:
+          java:
+            environment:
+              GOV_PAY_AUTH_KEY_CMC: la0ohpqa9er1bub2hddieeesc6sabbbqn4smkkve7s9ailn3n9d3hbnhc3
```
nitinprabhu, line 89: ?

k8s/demo/cluster-00/bsp/bsp-sample-app.yaml
```diff
@@ -0,0 +1,143 @@
+---
```
nitinprabhu, line 1: we don't need this file to be moved ?

PR #3352, opened 2020-05-06
Point api-gateway web directly to dm-store

timja:


hemantt:
@timja do we need the proxy to be added similarly to demo flux?

timja:
> @timja do we need the proxy to be added similarly to demo flux?

this is debugging for the P1 issue, dw about what we're doing here

PR #3357, opened 2020-05-06
Added ECM repos

jacarrete:
Added ECM repos


jacarrete:
> You need to add here as well https://github.com/hmcts/cnp-jenkins-config/blob/master/jobdsl/organisations-beta.groovy#L19

Added there as well https://github.com/hmcts/cnp-jenkins-config/pull/476

PR #3504, opened 2020-05-12
Trigger deployment of blob-router in demo

lgonczar:
### Change description ###

Trigger deployment of blob-router in demo

```
[ ] Yes
[x] No
```


k8s/demo/cluster-00/reform-scan/blob-router.yaml
```diff
@@ -26,6 +26,7 @@ spec:
         CRIME_ENABLED: "true"
         INCLUDE_WRITE_PERMISSION_IN_STORAGE_SAS_TOKEN: false
         FORCE_RESTART: "true"
+        REMOVE_ME: "dummy var to trigger deployment - remove it"
```
nitinprabhu, line 29: you could remove `FORCE_RESTART` above ?

PR #3510, opened 2020-05-13
Disable scheduled smoke tests for bulk-scan-processor in aat

lgonczar:
### Change description ###

Disable scheduled smoke tests for bulk-scan-processor in aat

```
[ ] Yes
[x] No
```


lgonczar:
> Any reason we are disabling this ?

Smoke are still configured to run after deplyments. I initially scheduled smoke tests to check if they work, but now I wanted to make them consistent with production smoke tests, where they don't run on schedule. I don't mind leaving it as it is.

PR #3593, opened 2020-05-18
Update jenkins req/limits

luigibk:
### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/namespaces/jenkins/patches/cftptl/cluster-00/jenkins.yaml
+++ b/k8s/namespaces/jenkins/patches/cftptl/cluster-00/jenkins.yaml
@@ -16,11 +16,11 @@ spec:
           value: https://cftptl-intsvc.vault.azure.net
       resources:
         requests:
-          cpu: "2"
-          memory: "8Gi"
-        limits:
           cpu: "4"
-          memory: "8Gi"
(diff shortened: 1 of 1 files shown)


k8s/namespaces/jenkins/patches/cftptl/cluster-00/jenkins.yaml
```diff
@@ -16,11 +16,11 @@ spec:
           value: https://cftptl-intsvc.vault.azure.net
       resources:
         requests:
-          cpu: "2"
-          memory: "8Gi"
-        limits:
           cpu: "4"
-          memory: "8Gi"
+          memory: "24Gi"
```
timja, line 20: not sure if it really needs this, could we maybe do 16 for a start?

luigibk: It's equivalent to Xmx=0.6*24GB = 14.4GB with the the rest of the RAM used by non-heap memory + other stuff, which jenkins seems to be happy to use. 
16GB as limit doesn't give us much heap memory (9.6) and it's a bit of a waste of these higher spec machines, but we can try...

PR #3711, opened 2020-05-21
Reduce SAS token permissions in blob-router prod

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-1187

### Change description ###

Remove `write` permission from SAS tokens served by blob-router in prod.

```
[ ] Yes
[x] No
```


lgonczar:
> Is it done for AAT ? If we want to enable it for all environments then best to update values yaml in router ?

This has been done in demo. I'll create another one for aat. When everything is fine, we'll completely get rid of the setting, but best to keep it for now, so that we can quickly revert it.

PR #3810, opened 2020-05-27
AAT-run post smoke and functional tests

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1231

### Change description ###

test configuration added 

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/aat/common/reform-scan/blob-router.yaml
+++ b/k8s/aat/common/reform-scan/blob-router.yaml
@@ -7,6 +7,15 @@ metadata:
   annotations:
     flux.weave.works/automated: "true"
     flux.weave.works/tag.java: glob:prod-*
+    fluxcd.io/automated: "true"
+    repository.fluxcd.io/java: java.image
+    filter.fluxcd.io/java: glob:prod-*
+    repository.fluxcd.io/functionalcron: java.functionaltestscron.image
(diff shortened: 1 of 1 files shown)


k8s/aat/common/reform-scan/blob-router.yaml
```diff
@@ -25,6 +34,43 @@ spec:
...
+              test-subscription-key: TEST_SUBSCRIPTION_KEY
+              storage-account-secondary-key: TEST_STORAGE_ACCOUNT_KEY
+              storage-account-name: TEST_STORAGE_ACCOUNT_NAME
+        environment:
+          TEST_URL: "http://reform-scan-blob-router-aat.service.core-compute-aat.internal"
+          SLACK_CHANNEL: "bsp-test-notices"
+          SLACK_NOTIFY_SUCCESS: true
+          TEST_STORAGE_CONTAINER_NAME: "bulkscan"
+          TEST_STORAGE_ACCOUNT_URL: "https://reformscanaat.blob.core.windows.net"
+          API_GATEWAY_URL: "https://core-api-mgmt-aat.azure-api.net/reform-scan"
```
doncem, line 62: should it not be platform url instead?

mdayican: it is for cert test. current url is like this in cnp jenkins

PR #3899, opened 2020-06-01
Remove an unused environment variable from blob-router's config

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-1187

### Change description ###

Remove a no longer used environment variable (`INCLUDE_WRITE_PERMISSION_IN_STORAGE_SAS_TOKEN`) from blob-router's config

```
[ ] Yes
[x] No
```


nitinprabhu:
> do we need change in Perftest ?



> do we need change in Perftest ?

didn't find env var in perftest so should be all good

PR #3993, opened 2020-06-04
Set SSCS Bulk Scanning urls as empty in prod

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1241

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/demo/cluster-00/bsp/bulk-scan-orchestrator.yaml
+++ b/k8s/demo/cluster-00/bsp/bulk-scan-orchestrator.yaml
@@ -15,7 +15,7 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: bulk-scan-orchestrator
-    version: 0.2.11
+    version: 0.2.12
   values:
     java:
       replicas: 2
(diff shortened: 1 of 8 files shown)


k8s/demo/cluster-00/bsp/bulk-scan-orchestrator.yaml
```diff
@@ -22,7 +22,6 @@ spec:
       useInterpodAntiAffinity: true
       image: hmctspublic.azurecr.io/bulk-scan/orchestrator:prod-c035f888
       environment:
-        TRANSFORMATION_URL_SSCS: http://sscs-bulk-scan-demo.service.core-compute-demo.internal
```
nitinprabhu, line 25: why do we need to remove it from demo ?

nitinprabhu: or your updating processor values yaml ?

aliveni: we don't need it after this change https://github.com/hmcts/bulk-scan-processor/pull/1328

aliveni: I thought of raising a separate PR for version. can do in this PR

nitinprabhu: so then we need chart bump in this PR. So first we need to merge PR 1328 and then this should have version bump change.
Also worth checking when is SSCS going to prod as we may just remove empty strings for prod

nitinprabhu: same goes for orchestrator

PR #4004, opened 2020-06-04
Extend send letter FTP downtime window to 6pm on prod

rkondratowicz:


rkondratowicz:
> do we need on perftest?

just prod

PR #4008, opened 2020-06-04
increase replica set for bsp - PROD

mdayican:
### Change description ###
increase replica set for bsp

```
[ ] Yes
[x] No
```


k8s/prod/common/bsp/bulk-scan-payment-processor.yaml
```diff
@@ -23,7 +23,7 @@ spec:
     version: 0.2.0
   values:
     java:
-      replicas: 2
+      replicas: 4
```
nitinprabhu, line 26: can be 2 :)

PR #4053, opened 2020-06-08
rename dev as preview, update readme

hannah38:


k8s/preview/cluster-00-overlay/kustomization.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: Kustomization
+bases:
+  - ../../namespaces/admin/flux
+  - ../../namespaces/admin/kube-slack
+  - ../../namespaces/admin/keyvault-flexvol
+  - ../../namespaces/admin/traefik
+  - ../../namespaces/kured
+  - ../../namespaces/monitoring
+  - ../../namespaces/admin/env-injector
+  - ../../namespaces/knode-system
```
adusumillipraveen, line 11: You can use osba as well from kustomize like in demo ?

adusumillipraveen: Same with keda

PR #4140, opened 2020-06-10
Remove fee keyword from ithc Divorce

smillapalk:
### Change description ###

As ithc is not used in Divorce, then removing the recently added fee keyword variable. 

**Does this PR introduce a breaking change?** 

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/ithc/common/divorce/div-fps.yaml
+++ b/k8s/ithc/common/divorce/div-fps.yaml
@@ -32,8 +32,6 @@ spec:
       aadIdentityName: divorce
       prometheus:
         enabled: true
-      environment:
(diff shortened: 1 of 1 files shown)


k8s/ithc/common/divorce/div-fps.yaml
```diff
@@ -32,8 +32,6 @@ spec:
       aadIdentityName: divorce
       prometheus:
         enabled: true
-      environment:
```
yarpo, line 35: don't we need it for ithc?

smillapalk: had a chat with Vivek and John and they said as we don't use them right now then adding this wouldn't add any value at this point of time

PR #4141, opened 2020-06-10
Feature toggle off FEATURE_PAPER_UPDATE on demo

yarpo:
Just to make sure bulk printing will not happen.

According to email I've got docs sent to bulk printing service on demo WILL BE printed and sent.

This is why I switch it off.  We can easily turn it on when needed (eg for acceptance testing).

### DIFF
--- a/k8s/demo/common/divorce/div-cos.yaml
+++ b/k8s/demo/common/divorce/div-cos.yaml
@@ -32,7 +32,7 @@ spec:
       environment:
         IDAM_API_URL: "https://idam-api.demo.platform.hmcts.net"
         FEATURE_DN_REFUSAL: "true"
-        FEATURE_PAPER_UPDATE: true
+        FEATURE_PAPER_UPDATE: false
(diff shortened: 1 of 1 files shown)


k8s/demo/common/divorce/div-cos.yaml
```diff
@@ -32,7 +32,7 @@ spec:
       environment:
         IDAM_API_URL: "https://idam-api.demo.platform.hmcts.net"
         FEATURE_DN_REFUSAL: "true"
-        FEATURE_PAPER_UPDATE: true
+        FEATURE_PAPER_UPDATE: false
```
jvictorcf, line 35: Can you write a comment explaining why this is off in demo, @yarpo ?

yarpo: According to what I've got from Leszek (BSP team) we don't need to feature toggle it off.

We need to provide them a uuid of doc uploaded and only these docs will be printed.

yarpo:
According to what I've got from Leszek (BSP team) we don't need to feature toggle it off.

We need to provide them a uuid of doc uploaded and only these docs will be printed.

PR #4312, opened 2020-06-19
Add flux config for AM Org role batch service

vijayrajagopalan-hmcts:
https://tools.hmcts.net/jira/browse/AM-819

### Change description ###

```
[ ] Yes
[X] No
```


k8s/aat/cluster-01/am/org-role-mapping-batch.yaml
```diff
@@ -0,0 +1,26 @@
...
+    version: 0.0.1
+  values:
+    job:
+      schedule: "*/10 * * * *"
+      useInterpodAntiAffinity: true
+      image: hmctspublic.azurecr.io/am/org-role-mapping-batch-service:prod-0cde3126
+    global:
+      environment: aat
+      tenantId: "531ff96d-0ae9-462a-8d2d-bec7c0b42082"
+      enableKeyVaults: true
```
luigibk, line 26: Isn't this exactly the same as the above one? Why don't you just add it to common instead?

vijayrajagopalan-hmcts: We are following the approach taken in another application.

Any documentation which explains when to use common and cluster-** will help.

k8s/aat/cluster-00/am/org-role-mapping-batch.yaml
```diff
@@ -0,0 +1,26 @@
...
+    version: 0.0.1
+  values:
+    job:
+      schedule: "*/10 * * * *"
+      useInterpodAntiAffinity: true
+      image: hmctspublic.azurecr.io/am/org-role-mapping-batch-service:prod-0cde3126
+    global:
+      environment: aat
+      tenantId: "531ff96d-0ae9-462a-8d2d-bec7c0b42082"
+      enableKeyVaults: true
```
luigibk, line 26: Are you sure this should run on a single cluster? I suggesetd moving to "common" not simply deleting one.

k8s/aat/cluster-01/am/org-role-mapping-batch.yaml
vijayrajagopalan-hmcts: Have removed config from cluster-01 as advised.

PR #4350, opened 2020-06-23
Add config for AM Judicial Booking Batch Service

vijayrajagopalan-hmcts:
https://tools.hmcts.net/jira/browse/AM-814

### Change description ###
Add config for AM Judicial Booking Batch Service

```
[ ] Yes
[X] No
```

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -228,6 +228,7 @@ k8s/aat/common/am/am-role-assignment-service.yaml @hmcts/am-devs
 k8s/aat/common/am/am-judicial-booking-service.yaml @hmcts/am-devs
 k8s/aat/cluster-**/am/role-assignment-batch.yaml @hmcts/am-devs
 k8s/aat/cluster-**/am/org-role-mapping-batch.yaml @hmcts/am-devs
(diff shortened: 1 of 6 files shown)


vijayrajagopalan-hmcts:
> Why is it on only one cluster ? Also, please add entry into codeowners

This is a batch job. At the moment it is not designed to manage parallel execution. When the application evolves, we may run it from both the clusters. Have added entry into codeowners

adusumillipraveen:
> > Why is it on only one cluster ? Also, please add entry into codeowners
> 
> This is a batch job. At the moment it is not designed to manage parallel execution. When the application evolves, we may run it from both the clusters. Have added entry into codeowners

Add it at a cron schedule where it doesn't clash with each other (that's what most of the teams do) . Also, it's not good to design new jobs in this way.

vijayrajagopalan-hmcts:
> > > Why is it on only one cluster ? Also, please add entry into codeowners
> > 
> > 
> > This is a batch job. At the moment it is not designed to manage parallel execution. When the application evolves, we may run it from both the clusters. Have added entry into codeowners
> 
> Add it at a cron schedule where it doesn't clash with each other (that's what most of the teams do) . Also, it's not good to design new jobs in this way.

Have added the jobs on cluster-01 and set to run at different times.

PR #4353, opened 2020-06-23
feat: remove references to CCD UI from FPL when live(FPLA-1772)

KiranVarma02:

https://tools.hmcts.net/jira/browse/FPLA-1772

### Change description ###

Remove references to CCD UI from FPL when live

```
[ ] Yes
[ *] No
```

### DIFF
--- a/k8s/aat/common/family-public-law/fpl-case-service.yaml
+++ b/k8s/aat/common/family-public-law/fpl-case-service.yaml
@@ -15,7 +15,7 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: fpl-case-service
-    version: 1.11.8
+    version: 1.11.7
   values:
     java:
       replicas: 2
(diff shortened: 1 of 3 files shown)


k8s/aat/common/family-public-law/fpl-case-service.yaml
```diff
@@ -15,7 +15,7 @@ spec:
   chart:
     repository: https://hmctspublic.azurecr.io/helm/v1/repo/
     name: fpl-case-service
-    version: 1.11.4
+    version: 1.11.7
```
lewisbirks, line 18: should this not be .8? same in others as well

KiranVarma02: oh, its moved to 8 now..it was 7 when i created PR :)

KiranVarma02: This PR is not going to be merged until after expert UI is live. So i guess i will make this change later.

tomaszpowroznik: it basically should match version in your fpl-ccd-configuration Chart.yaml which currently is .8

PR #4380, opened 2020-06-24
RPET-238: Add payment keyword to FR prod

smillapalk:
### JIRA link ###

https://tools.hmcts.net/jira/browse/RPET-238

### Change description ###

Changing General Application Without Notice fee keyword on prod for FR - coordinated with FeePay. 
As flux was used to merge per environment, this is now deleted. 
The value is instead stored in finrem-payments-service in values.yaml as all environments will from this PR onwards use the same keyword. 

**Does this PR introduce a breaking change?**

```
[ ] Yes
[x] No
```


Cotterale:
Re-opening, Fee keyword update is going ahead

PR #4424, opened 2020-06-26
increase memory

mdayican:
### Change description ###

increase memory to see would it work in  aat-01

```
[ ] Yes
[x] No
```


k8s/aat/common/reform-scan/blob-router.yaml
```diff
@@ -26,6 +26,7 @@ spec:
   values:
     java:
       replicas: 2
+      memoryRequests: "2048Mi"
```
nitinprabhu, line 29: 2048 is by default as per chart java

PR #4531, opened 2020-07-02
Lower log level in aat/demo for bsp payments

doncem:

[Log the content of CCD error responses in non-production environments](https://tools.hmcts.net/jira/browse/BPS-697)

### Change description ###

There's a single call in payment processor - amending level for those environments too. Will only take into affect once

- hmcts/bulk-scan-payment-processor#242
- future pr to bump logging version to 5.1.6 is merged

```
[ ] Yes
[x] No
```


doncem:
turned out it is not needed

PR #4611, opened 2020-07-07
Added in prod file for ICP

yogesh-hullatti:
https://tools.hmcts.net/jira/browse/EM-3020

### DIFF
--- a/k8s/prod/cluster-00/em/em-icp.yaml
+++ b/k8s/prod/cluster-00/em/em-icp.yaml
@@ -1,0 +1,30 @@
+---
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
+  name: em-icp
+  namespace: em
+  annotations:
+    flux.weave.works/automated: "true"
+    flux.weave.works/tag.nodejs: glob:prod-*
+spec:
+  releaseName: em-icp
+  rollback:
+    enable: true
+    retry: true
+  chart:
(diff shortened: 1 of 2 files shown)


k8s/prod/cluster-00/em/em-icp.yaml
```diff
@@ -0,0 +1,30 @@
...
+  chart:
+    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
+    name: em-icp
+    version: 0.1.10
+  values:
+    nodejs:
+      replicas: 1
+      useInterpodAntiAffinity: true
+      image: hmctspublic.azurecr.io/em/icp:prod-ae3d2e63
+      ingressHost: em-icp-prod.service.core-compute-prod.internal
```
timja, line 24: This should be in your chart

timja, line 24: This should be in your chart

k8s/prod/cluster-00/em/em-icp.yaml
```diff
@@ -0,0 +1,30 @@
...
+      environment:
+        VAR_T3: trigger4
```
timja, line 26: This shouldn’t be needed

yogesh-hullatti: Thank you @timja . I will need to implement this across all our API's. Will implement this review comment as part of separate ticket.

PR #4612, opened 2020-07-07
Enable processing notifications queue messages

aliveni:

```
[ ] Yes
[x] No
```


aliveni:
> enable now or when duplicate handling code is present in source code?

I'll wait for the duplicate handling PR

PR #4617, opened 2020-07-07
SSCS-6822 Add TYA Link for AAT

danhamilton:

- [X] commit messages are meaningful and follow good commit message guidelines
- [X] README and other documentation has been updated / added (if needed)
- [X] tests have been updated / new tests has been added (if needed)

https://tools.hmcts.net/jira/browse/SSCS-6822

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/k8s/aat/common/sscs/sscs-tya-notif.yaml
+++ b/k8s/aat/common/sscs/sscs-tya-notif.yaml
@@ -25,6 +25,7 @@ spec:
         JOB_SCHEDULER_DB_USERNAME: notification@sscs-tya-notif-postgres-v11-db-aat
         JOB_SCHEDULER_DB_NAME: notification
         JOB_SCHEDULER_DB_HOST: sscs-tya-notif-postgres-v11-db-aat.postgres.database.azure.com
+        SSCS_TRACK_YOUR_APPEAL_LINK: https://track-appeal.aat.platform.hmcts.net/trackyourappeal/appeal_id
     global:
       environment: aat
       tenantId: "531ff96d-0ae9-462a-8d2d-bec7c0b42082"


k8s/aat/common/sscs/sscs-tya-notif.yaml
```diff
@@ -25,6 +25,7 @@ spec:
         JOB_SCHEDULER_DB_USERNAME: notification@sscs-tya-notif-postgres-v11-db-aat
         JOB_SCHEDULER_DB_NAME: notification
         JOB_SCHEDULER_DB_HOST: sscs-tya-notif-postgres-v11-db-aat.postgres.database.azure.com
+        SSCS_TRACK_YOUR_APPEAL_LINK: https://track-appeal.aat.platform.hmcts.net/trackyourappeal/appeal_id
```
ryandac, line 28: Is this the correct place to add this change https://github.com/hmcts/sscs-track-your-appeal-notifications/blob/master/charts/sscs-tya-notif/values.yaml#L44 ?

danhamilton: Yes I think so as:

- It's equivalent to how the PROD value is used which is important for the test.
- It's a AAT specific value, and not the default value to be injected.

Aside: I think it is right we are doubly vigilant re config changes, from what I can see our BE services will fallback to AAT or other values in PROD if conf not correctly set; where I think it's safer the app fails to start and deployment fails with existing running pods still up.

ryandac: Cool, as long as that change in notifications is not lost.

PR #4650, opened 2020-07-08
Jenkins GC + pod affinity/antiaffinity

luigibk:

[[AB#867](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/867)]

### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/namespaces/jenkins/patches/cftptl/cluster-00/jenkins.yaml
+++ b/k8s/namespaces/jenkins/patches/cftptl/cluster-00/jenkins.yaml
@@ -17,10 +17,10 @@ spec:
       resources:
         requests:
           cpu: "4"
-          memory: "20Gi"
+          memory: "24Gi"
         limits:
           cpu: "6"
-          memory: "20Gi"
+          memory: "24Gi"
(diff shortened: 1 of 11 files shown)


k8s/namespaces/jenkins/patches/cftptl/cluster-00/jenkins.yaml
```diff
@@ -17,10 +17,10 @@ spec:
       resources:
         requests:
           cpu: "4"
-          memory: "20Gi"
+          memory: "24Gi"
         limits:
           cpu: "6"
-          memory: "20Gi"
+          memory: "24Gi"
```
timja, line 23: Didn’t we try 24 once and there wasn’t enough memory?

luigibk: Yes, it’s a bit of a gamble, but if those other workloads stay away from the node it should work as there are currently 28GB allocatable

PR #4916, opened 2020-07-22
Sandbox sealed secrets for acr-credentials instead of mount

hannah38:

```
[ ] Yes
[ x] No
```


hannah38:
> Why not in common ?

Basically in azure.json there were some differences between the values in cluster 00 and 01. Like for subnetName and resourceGroup

timja:
> > Why not in common ?
> 
> Basically in azure.json there were some differences between the values in cluster 00 and 01. Like for subnetName and resourceGroup

they shouldn't be relevant for flux

PR #5042, opened 2020-07-28
Revert "Scale services for running on one cluster"

timja:
Reverts hmcts/cnp-flux-config#5029


timja:
This is just reverting a change from a scale up this morning for maintenance

PR #5077, opened 2020-07-29
DIV-6196: to fix prod bug - update image on prod to prod-dc4127d0

yarpo:
We need to do manually
Screenshot 2020-07-29 15 40 39

### DIFF
--- a/k8s/prod/common/divorce/div-cos.yaml
+++ b/k8s/prod/common/divorce/div-cos.yaml
@@ -28,7 +28,7 @@ spec:
       readinessPeriod: 15
       useInterpodAntiAffinity: true
       applicationInsightsInstrumentKey: "baaed5d0-7666-4dce-bf74-edfceeb566f9"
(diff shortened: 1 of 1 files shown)


k8s/prod/common/divorce/div-cos.yaml
```diff
@@ -28,7 +28,7 @@ spec:
       readinessPeriod: 15
       useInterpodAntiAffinity: true
       applicationInsightsInstrumentKey: "baaed5d0-7666-4dce-bf74-edfceeb566f9"
-      image: hmctspublic.azurecr.io/div/cos:prod-a25c3b5d
+      image: hmctspublic.azurecr.io/div/cos:prod-dc4127d0
```
yarpo, line 31: demo is updated (autopmatically).
aat is using `prod-*`.

yarpo: image updated on aat:
```
   prod-dc4127d0
```

PR #5197, opened 2020-08-04
Change PR link for `bulk-scan-orchestrator` in demo

doncem:
### Change description ###

Switching deployment flag so manual activity in orchestrator can be verified against automated models

```
[ ] Yes
[x] No
```


doncem:
thought that updates automatically by flux. the main thing what needs is that flag `automated = true` and regex for image deployments ;). flux should update line:23 automatically then

not the case?

nitinprabhu:
> thought that updates automatically by flux. the main thing what needs is that flag `automated = true` and regex for image deployments ;). flux should update line:23 automatically then
> 
> not the case?

nopes since you are using specific image of PR

PR #5231, opened 2020-08-05
Sandbox aad-pod-identity

hannah38:
Chart version is actually 2.0.1 for app v 1.6.2, so needing fixing.
Also enabled rollbacks and install of CRDs.

```
[ ] Yes
[ x] No
```


k8s/namespaces/admin/aad-pod-identity/aad-pod-identity.yaml
```diff
@@ -6,11 +6,17 @@ metadata:
     k8s-app: aad-pod-id
   name: aad-pod-identity
   namespace: admin
+  annotations:
+    fluxcd.io/automated: "false"
```
timja, line 10: is this needed?

PR #5240, opened 2020-08-05
Add ingress host wa workflow api

chrisgrimble:
Add ingress host wa workflow api

### DIFF
--- a/k8s/namespaces/wa/wa-workflow-api/wa-workflow-api.yaml
+++ b/k8s/namespaces/wa/wa-workflow-api/wa-workflow-api.yaml
@@ -12,4 +12,4 @@ spec:
     java:
       replicas: 2
       useInterpodAntiAffinity: true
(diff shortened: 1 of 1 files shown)


k8s/namespaces/wa/wa-workflow-api/wa-workflow-api.yaml
```diff
@@ -13,3 +13,4 @@ spec:
       replicas: 2
       useInterpodAntiAffinity: true
       image: hmctspublic.azurecr.io/wa-workflow-api/service:prod-8431c386
+      ingressHost: wa-workflow-api.service.core-compute-aat.internal
```
raghera, line 16: think should be something like: 
`ingressHost: wa-workflow-api-aat.service.core-compute-aat.internal`
e.g. ia-case-api-aat.service.core-compute-aat.internal/

PR #5319, opened 2020-08-07
Am 959 ras aat demo

vijayrajagopalan-hmcts:
https://tools.hmcts.net/jira/browse/AM-959

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/k8s/demo/common-overlay/am/kustomization.yaml
+++ b/k8s/demo/common-overlay/am/kustomization.yaml
@@ -3,3 +3,6 @@ kind: Kustomization
 namespace: am
 bases:
 - ../../../namespaces/am
+- ../../../namespaces/am/am-role-assignment-service/am-role-assignment-service.yaml
+patchesStrategicMerge:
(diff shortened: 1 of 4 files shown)


k8s/demo/common-overlay/am/kustomization.yaml
```diff
@@ -3,3 +3,6 @@ kind: Kustomization
 namespace: am
 bases:
 - ../../../namespaces/am
+- ../../../namespaces/am/am-role-assignment-service/am-role-assignment-service.yaml
```
timja, line 6: I don't think this is quite right, you've added it to the global kustomization and the demo one? don't you only need to add it to one?

vijayrajagopalan-hmcts: Global as in k8s/namespaces/am/kustomization.yaml ? I have not added here. I want this service to be deployed only on AAT and Demo for now.

PR #5333, opened 2020-08-07
increase memory and cpu for definition store

MSancaktutar:
--

### Change description ###
increase memory and cpu for definition store

```
[ ] Yes
[X] No
```


hemantt:
Please can we get an approval from PE team @MSancaktutar..
Just that this is related to the resource allocations?

MSancaktutar:
@timja - Do you have any concerns with this change? We had applied this to mitigate the performance issues on AAT. The performance has actually been improved by an implementation change, but if it's not a big issue, I'd prefer these to stay like what's been merged. Do you think we should rather adjust these to different values?

PR #5384, opened 2020-08-11
bulk-scan-orchestrator restart aat pods

mdayican:
### Change description ###

bulk-scan-orchestrator restart aat pods

```
[ ] Yes
[x] No
```


mdayican:
> Reason would be nice why pods needs to be restarted :)

In aat-00 pods are not recreated after helm delete

PR #5512, opened 2020-08-18
SIDM-4641 Add a new env variable to be injected into web public.

radoslaw-orlowski-amido:
### Change description ###

Add the SSO feature flag to prod config.

```
[ ] Yes
[x] No
```


henrydobson (changes requested):
I presume you're trying to set this early because the default is SSO enabled? If so, this change should be bundled in with the changes for 2.2.

radoslaw-orlowski-amido:
To be added to 2.2 by @henrydobson

PR #5659, opened 2020-08-24
XUI update flag

ernestman28:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/k8s/ithc/common/xui/webapp.yaml
+++ b/k8s/ithc/common/xui/webapp.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: xui
   annotations:
     fluxcd.io/automated: "true"
-    fluxcd.io/tag.nodejs: glob:pr-587-66ffed21
+    fluxcd.io/tag.nodejs: glob:pr-616-8ff52e1a
 spec:
   releaseName: xui-webapp
   rollback:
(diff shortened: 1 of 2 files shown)


k8s/ithc/common/xui/webapp.yaml
```diff
@@ -29,8 +29,11 @@ spec:
         LAUNCH_DARKLY_CLIENT_ID: 1
         DEBUG: xuiNode:*,-xuiNode:auth:s2s
         ISS_SERVICE: https://forgerock-am.service.core-compute-idam-ithc.internal:8443/openam/oauth2/hmcts
-        SERVICES_IDAM_OAUTH_CALLBACK_URL: /oauth2/callback
+        SERVICES_IDAM_OAUTH_CALLBACK_URL: oauth2/callback
```
udaydenduluri33, line 32: Why are you removing this /

ernestman28: same setting as MC demo

ernestman28: this is currently in MC demo which is working right, if we going to use the same tag, this env var will need to use the same too? @udaydenduluri33 https://github.com/hmcts/cnp-flux-config/blob/master/k8s/demo/common/xui/webapp.yaml#L33

udaydenduluri33: If using the new node lib, then it's /ouath2/callback

ernestman28: we using exact same tagged image as demo, so env var should be the same right? until we get a new PR, i think thats should be correct

PR #5972, opened 2020-09-15
enable reconciliation report in AAT

mdayican:
### Change description ###

enable reconciliation report in AAT

```
[ ] Yes
[x] No
```


doncem (approved):
..but we won't need this to be running on aat, right?

mdayican:
> ..but we won't need this to be running on aat, right?

For QA testing

apereverzin (approved):

PR #6022, opened 2020-09-16
Add test config for PCQ

aliveni:

```
[ ] Yes
[x] No
```


doncem (approved):

mdayican (approved):
if you are not use these in test it would not fix but we can try.

PR #6044, opened 2020-09-16
correct identity

iguannago:
https://tools.hmcts.net/jira/browse/RWA-119


hannah38 (changes requested):
Please update the client ID as well

hannah38 (approved):

PR #6051, opened 2020-09-17
Disable automatic case creation for probate in demo

aliveni:

```
[ ] Yes
[x] No
```


doncem (approved):
i presume defaults are false ;)

aliveni:
> i presume defaults are false ;)

Yes, they are false :)

PR #6055, opened 2020-09-17
Switched aad-pod-identity cfg to raw manifest

gbstan92:
Switch sbox01 aad-pod-identity cfg to raw manifest

### DIFF
--- a/k8s/sandbox/cluster-01-overlay/kustomization.yaml
+++ b/k8s/sandbox/cluster-01-overlay/kustomization.yaml
@@ -32,10 +32,7 @@ patchesStrategicMerge:
 
   #kured patch
   - ../../namespaces/kured/patches/sandbox/cluster-01/kured.yaml
-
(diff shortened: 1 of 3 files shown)


k8s/sandbox/cluster-01-overlay/kustomization.yaml
```diff
@@ -33,6 +33,9 @@ patchesStrategicMerge:
   #kured patch
   - ../../namespaces/kured/patches/sandbox/cluster-01/kured.yaml
 
+  #aad-pod-identity patch
```
adusumillipraveen, line 36: now, i am thinking this patch should just go into base kustomize ? as its same across all clusters

gbstan92: resolved

adusumillipraveen (approved):

luigibk (approved):

PR #6088, opened 2020-09-21
Add MI

linusnorton:
Add fact MI to perftest and aat


hannah38 (commented):
This looks good, however you might just want to keep the name, azureidentity and selector to "fact" as this will cover more than just the api pods

nigeldunne (approved):

PR #6120, opened 2020-09-22
Add config to apply demo patch deployment

iguannago:
https://tools.hmcts.net/jira/browse/RWA-122


k8s/namespaces/ia/case-api/demo.yaml
```diff
@@ -0,0 +1,34 @@
+---
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
+  name: ia-case-api
+  namespace: ia
+  annotations:
+    flux.weave.works/automated: "true"
+    flux.weave.works/tag.java: glob:wa-proto-demo-*
```
adusumillipraveen, line 9: this is not a valid glob. Also patch should only have the things you want apply on top of original manifest.

adusumillipraveen (changes requested):

k8s/namespaces/ia/case-api/demo.yaml
iguannago: Thanks, @adusumillipraveen, for your feedback. I have fixed those issues. Please, could you confirm it is OK now? Thanks. 
One more thing, please, DO NOT MERGE since some other team is using Demo and I don't want to deploy my changes just yet. Thanks.

iguannago:
Wrong configuration. I open this new PR instead https://github.com/hmcts/cnp-flux-config/pull/6131

PR #6358, opened 2020-10-07
Add aat vault to preview

linusnorton:


nigeldunne (dismissed):

hannah38 (changes requested):
Preview is slightly different (using the aat mi), so can you change this to - ../../../aat/common/fact/identity.yaml

hannah38 (approved):

PR #6601, opened 2020-10-21
Jenkins VM Agent Initscript to exit on non-zero exit code

gbstan92:


avi-hmcts (approved):

k8s/namespaces/jenkins/patches/cftptl/cluster-00/jenkins.yaml
```diff
@@ -414,6 +414,7 @@ spec:
                           mkdir -pv /opt/jenkins
                           mount /dev/sdb1 /opt/jenkins
                           chown -R jenkinsssh:jenkinsssh /opt/jenkins
+                          set -e
```
timja, line 417: Any reason to not do this for the whole script?

avi-hmcts: as we are seeing errors when doing a `umount /mnt/resource` _set -e_ would probably not allow us to actually check the main issue `cp /tmp/jenkinsssh_id_rsa /home/jenkinsssh/.ssh/id_rsa`

timja: Prob should get rid of that then

PR #6634, opened 2020-10-26
Disable api auth temporarily till teams migrate

adusumillipraveen:

```
[ ] Yes
[x] No
```


timja (approved):
would be good to have it enabled somewhere? maybe preview?

adusumillipraveen:
> would be good to have it enabled somewhere? maybe preview?

They are using Camunda chart in preview , so it comes on by default when they upgrade the chart.

PR #7027, opened 2020-11-17
Remove secret reference "fail-secret"

thomast1906:
### Change description ###
Removing reference to "fail-secret" as it is not a valid secret name in the KeyVault.

```
[ ] Yes
[X] No
```


msl8r (approved):
👍

adusumillipraveen (approved):

PR #7108, opened 2020-11-23
[WIP]RPET-491: Add HelmRelease that should refer to the charts automatically publish…

jacmurphy:
https://tools.hmcts.net/jira/browse/RPET-491

### Change description ###

Currently, we need to upgrade our flux files for every new version of our helm charts.

There is a solution, however, to make this manual (and error-prone) process redundant, which is to make flux use GitHub charts instead of ACR.

It seems to be a relatively easy fix with a quite high value.

```
[ ] Yes
[ ] No
```

### DIFF
--- a/k8s/namespaces/divorce/namespace.yaml
+++ b/k8s/namespaces/divorce/namespace.yaml
@@ -5,3 +5,8 @@ metadata:
   labels:
     slackChannel: div-dev
     hmcts.github.com/envInjector: enabled
+spec:
+  chart:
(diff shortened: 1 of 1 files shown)


k8s/namespaces/divorce/namespace.yaml
```diff
@@ -5,3 +5,8 @@ metadata:
   labels:
     slackChannel: div-dev
     hmcts.github.com/envInjector: enabled
+spec:
```
yarpo, line 8: why?

PR #7297, opened 2020-12-03
Push demo build

apereverzin:
### Change description ###

```
[ ] Yes
[x] No
```


mdayican (commented):
There is a problem in demo cluster , it can not bind key vault to pod so new pods  can not run.

PR #7315, opened 2020-12-04
use branch of bsp router on perftest

jasonpaige:

```
[ ] Yes
[X] No
```


k8s/perftest/common-overlay/bsp/kustomization.yaml
```diff
@@ -4,5 +4,6 @@ namespace: bsp
 bases:
   - ../../../namespaces/bsp
 patchesStrategicMerge:
+  - ../../../namespaces/bsp/blob-router-service/perftest.yaml
```
mdayican, line 7: blob router is under reform-scan namespace, it should be there

jasonpaige: :facepalm:

mdayican (approved):

PR #7523, opened 2020-12-29
Em 3318 add em hrs api

yogesh-hullatti:
https://tools.hmcts.net/jira/browse/EM-3318


adusumillipraveen (changes requested):
All new apps should follow this.

yogesh-hullatti:
NOt required

PR #7977, opened 2021-02-03
Added in files to override var on each environment

yogesh-hullatti:
Added in files to override var on each environment


adusumillipraveen (commented):
why not do it when u need it ?

benouaer (approved):

yogesh-hullatti:
> why not do it when u need it ?

Getting ready now. As we have seen it takes long to get the PR approved by DTSPO team. And it will be quick for us on teh day of migration to just update existing file and get it reviewed and approved quickly with in our team. Also. The migration activity  will be planned out of office hours. Hope it makes sense. We plan to do similar activity for other 3 EM API's.

timja (approved):
doesn't cause any harm

PR #8229, opened 2021-02-16
Adding apps to common kustomization

raghera:
Adding standalong-task-bpmn and case-event-handler to common kustomization and removing from the AAT/Demo common-overlay since we want the apps to be deployed everywhere.


alexbaeza (approved):
LGTM 👀

adusumillipraveen (commented):
Do they have Authority to operate on prod ?

adusumillipraveen (approved):

PR #8302, opened 2021-02-19
DIV-6800: Disabling SaC for test

jvictorcf:
### Change description ###
I don't think that's the issue, but I'll try to bypass the SaC logic to see if the bug still happens.

```
[] Yes
[] No
```


yarpo (approved):
why?

jvictorcf:
> why?

AllowShareACaseTask is failing
I don't think that's causing the issue, but I want to rule it out for sure.
I'll revert this soon after the test

PR #8316, opened 2021-02-22
Adding docmosis prod

adusumillipraveen:

```
[ ] Yes
[x] No
```


hannah38 (dismissed):

k8s/namespaces/docmosis/docmosis/prod.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
+  name: docmosis
+  annotations:
+    filter.fluxcd.io/chart-image: glob:prod-*
+spec:
+  values:
+    ingressHost: docmosis.platform.hmcts.net
```
timja, line 9: there's also a .prod.platform.hmcts.net domain, it might be better to use that? 
it's meant to be used so that people don't have to override prod domains to remove the .environment

timja (dismissed):

k8s/namespaces/docmosis/docmosis/prod.yaml
adusumillipraveen: But then this migration will depend on teams migrating to the latest config which i want to avoid.

timja (approved):

PR #8749, opened 2021-03-18
restart UP pods on all env

shreedhar-hmcts:
restart UP pods on all env like prod , aat, demo, ithc, perftest


pksaxena01 (approved):
Approved

PR #8750, opened 2021-03-18
restart PRD pods

shreedhar-hmcts:
restart PRD pods on all env like prod , aat, demo, ithc, perftest


shashank-rastogi (approved):

pksaxena01 (approved):
Approved

PR #8823, opened 2021-03-25
DTSPO-1452 Demo ENV: migrating to sendgrid  

mokainos:
https://tools.hmcts.net/jira/browse/DTSPO-1452


timja (changes requested):
no need for this just do it in the chart in the github repo and roll it out to the rest of the envs

PR #8846, opened 2021-03-26
Permissions for teams to del pods and hr on nonprod

hannah38:
Extra permissions for teams to delete pods and hr in their namespaces.
For ITHC and idam on sandbox so far.
Two groups I still need to find in AAD - for fact and nfdiv teams.

```
[ ] Yes
[ x] No
```


timja:
@hannah38 nfdiv uses the divorce group

adusumillipraveen (approved):

PR #9066, opened 2021-04-08
RWA-474  disable automated prod image deployments

iguannago:
https://tools.hmcts.net/jira/browse/RWA-474


njrich28 (dismissed):
Approved to override prod images in ITHC for purposes of pen testing

k8s/ithc/common-overlay/wa/kustomization.yaml
```diff
@@ -6,6 +6,10 @@ bases:
 - ../../../namespaces/all-namespaces/nonprod-role.yaml
 patchesStrategicMerge:
 - ../../../namespaces/wa/wa-case-event-handler/ithc.yaml
+- ../../../namespaces/wa/wa-standalone-task-bpmn/ithc.yaml
+- ../../../namespaces/wa/wa-task-configuration-api/ithc.yaml
+- ../../../namespaces/wa/wa-task-management-api/ithc.yaml
+- ../../../namespaces/wa/wa-workflow-api/ithc.yaml
```
raghera, line 12: can't see it in the PR diff but I am assuming the `ia-task-configuration` kustomization ref is there already?

k8s/namespaces/ia/ia-task-configuration/ithc.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
+  name: ia-task-configuration
+  annotations:
+    fluxcd.io/automated: "true"
+    fluxcd.io/tag.java: glob:pr-85-*
```
raghera, line 7: Do we have to tag a PR here?  Could we have not just done something like
`fluxcd.io/automated: "true"`

If it is the same effect then it's fine to keep it as it is.

k8s/namespaces/ia/ia-task-configuration/ithc.yaml
```diff
@@ -0,0 +1,12 @@
...
+    hmcts.github.com/prod-automated: disabled
+spec:
+  values:
+    java:
+      image: hmctspublic.azurecr.io/ia/ia-task-configuration:prod-3bd0b248
```
raghera, line 12: I can't remember but I guess Flux will replace this with the PR tag when the operator runs right?

k8s/ithc/common-overlay/wa/kustomization.yaml
iguannago: yes, I added too. Here: https://github.com/hmcts/cnp-flux-config/pull/9066/files#diff-6038c73edc3f2dd661218d6caab98dcf36817411d8fb967b1977527c6aa83dd7R9 

and here: https://github.com/hmcts/cnp-flux-config/pull/9066/files#diff-dc949e42add5d629316a1021b43a1f12eb21af83a7ad075af2cd106f6a6654f1R1

k8s/namespaces/ia/ia-task-configuration/ithc.yaml
iguannago: by tagging the PR we can use this PR  to deploy any feature we might need to deploy manually.  In any case, I am not sure it would work without it. This is how it's documented and platops approved it too.

k8s/namespaces/ia/ia-task-configuration/ithc.yaml
iguannago: Correct. It just needs to exist. Otherwise, it will fail.

k8s/namespaces/ia/ia-task-configuration/ithc.yaml
raghera: Makes sense actually so thanks, good to know!

k8s/namespaces/ia/ia-task-configuration/ithc.yaml
raghera: nice one

raghera (approved):

raghera (approved):

k8s/namespaces/ia/kustomization.yaml
```diff
@@ -5,3 +5,4 @@ bases:
 - namespace.yaml
 - ia-home-office-integration-api/ia-home-office-integration-api.yaml
 - ia-case-payments-api/ia-case-payments-api.yaml
+- ia-task-configuration/ia-task-configuration.yaml
```
timja, line 8: this change adds it to all environments, is that what is intended?

timja (dismissed):

timja (approved):

BenNzewi (approved):

PR #9078, opened 2021-04-08
DTSPO-1875 - fix artifactory config load

msl8r:
### Change description ###
- Config file doesn't load properly, using artifactory.config.import.xml seems to work even when there is existing configuration in place. 
- Testing with a patch, once I know it works I will move the new image and command back into the main artifactory file
- Fix linting warnings 

```
[ ] Yes
[x] No
```


timja (commented):
shouldn't you just be patching what you need to change? not everything?

msl8r:
> shouldn't you just be patching what you need to change? not everything?

I just did it for speed as it's a temporary change, just wanted to test that this works

timja (dismissed):

msl8r:
@timja - i've updated, how's that?

timja (approved):

PR #9079, opened 2021-04-08
Bug Fix - Adding Sealed Secret Hack for AAT

CathMcK:
### Change description ###

Bug Fix - Adding Sealed Secret Hack for AAT

```
[ ] Yes
[ X] No
```


adusumillipraveen (approved):

k8s/aat/common/neuvector/nv-storage-secret-hack.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: bitnami.com/v1alpha1
+kind: SealedSecret
+metadata:
+  creationTimestamp: null
+  name: nv-storage-secret
+  namespace: default
```
timja, line 6: should really add a comment here linking to why this is in the default namespace

CathMcK: Hi Tim, Its a bug in aks 1.19.7 - https://github.com/kubernetes/kubernetes/issues/99061 
This is a workaround until we can upgrade to the next version which contains the fix.

timja: i'm aware but plenty of others aren't

PR #9140, opened 2021-04-13
Update sandbox to match preview

verityit:
Updating sandbox to use preview images
updated chart version

```
[ ] Yes
[] No
[X] Hope not !
```

### DIFF
--- a/k8s/sandbox/common/idam/idam-api.yaml
+++ b/k8s/sandbox/common/idam/idam-api.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: idam
   annotations:
     flux.weave.works/automated: "true"
-    flux.weave.works/tag.java: glob:aat-*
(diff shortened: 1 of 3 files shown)


k8s/sandbox/common/idam/idam-api.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   namespace: idam
   annotations:
     flux.weave.works/automated: "true"
```
jburke-idam, line 8: you might want to turn this off unless you want to pick up other preview changes as they come alone (if flux worked)

jburke-idam (approved):

k8s/sandbox/common/idam/idam-api.yaml
verityit: i've specified an image name, so update to preview wouldnt deploy to sandbox ?

PR #9160, opened 2021-04-14
RDM-10969: Migrate Data store DB and Logstash for demo env

kiran-yenigala-hmcts:

https://tools.hmcts.net/jira/browse/RDM-10969 

### Change description ###

Message Publisher changes

```
[ ] Yes
[X ] No
```

### DIFF
--- a/k8s/demo/cluster-00-overlay/kustomization.yaml
+++ b/k8s/demo/cluster-00-overlay/kustomization.yaml
@@ -30,7 +30,6 @@ bases:
 
 patchesStrategicMerge:
   #flux patches
-  - ../../../namespaces/ccd/message-publisher/demo.yaml
   - ../../namespaces/admin/flux/patches/demo/flux.yaml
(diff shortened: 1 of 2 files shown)


k8s/demo/cluster-00-overlay/kustomization.yaml
```diff
@@ -30,6 +30,7 @@ bases:
 
 patchesStrategicMerge:
   #flux patches
+  - ../../../namespaces/ccd/message-publisher/demo.yaml
```
danlysiak, line 33:
```suggestion
```
Since Demo is only one cluster we already have the generic demo.yaml applied to the common-overlay (see [here](https://github.com/hmcts/cnp-flux-config/blob/master/k8s/demo/common-overlay/ccd/kustomization.yaml#L9)), so no need to update this file

danlysiak (approved):

PR #9170, opened 2021-04-14
DTSPO-2296 - Payment Sendgrid migration

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-2296

### Change description ###
Adding key vault and environment variables for migrating Payment AAT to SendGrid

```
[ ] Yes
[ x ] No
```


timja (changes requested):
Secrets should be using the values from the helm chart (no .'s are allowed in secrets)

https://github.com/hmcts/ccpay-payment-app/blob/master/charts/payment-api/values.yaml#L83-L123

endakelly:
> Secrets should be using the values from the helm chart (no .'s are allowed in secrets)
> 
> https://github.com/hmcts/ccpay-payment-app/blob/master/charts/payment-api/values.yaml#L83-L123

Fixed as suggested by Tim

timja (changes requested):
the actual password seems to be missing?

`- spring-mail-password` unless I'm missing something?

endakelly:
> the actual password seems to be missing?
> 
> `- spring-mail-password` unless I'm missing something?

Yeah I missed that one. Added in there now

timja (approved):

PR #9191, opened 2021-04-15
chore(ci): add aac identity to fpl preview

lewisbirks:
Attempting to fix the following:
```
❯ k describe pod fpl-case-service-pr-2505-aac-manage-case-assignment-7445f5bx8cq

...

  Warning  FailedMount  11m   kubelet            MountVolume.SetUp failed for volume "vault-aac" : rpc error: code = Unknown desc = error mounting secret time="2021-04-15T12:06:19Z" level=fatal msg="[error] : failed to get keyvaultClient: failed to get key vault token: nmi response failed with status code: 404"

```

Previous attempt: https://github.com/hmcts/cnp-flux-config/pull/9187
FPL chart setup: https://github.com/hmcts/fpl-ccd-configuration/pull/2505


adusumillipraveen (commented):
could you use same approach for other identities in https://github.com/hmcts/cnp-flux-config/tree/7f3efc496a7bf81a77572ba8987e16a91e971c11/k8s/preview/common/family-public-law pls

lewisbirks:
@adusumillipraveen removed xui, not sure what to do about the idam one. I am not really sure what I am doing and am just following suggestions.

adusumillipraveen:
> @adusumillipraveen removed xui, not sure what to do about the idam one. I am not really sure what I am doing and am just following suggestions.

you are just referring to Manage Identity files already used by other teams to access AAT keyvaults instead of duplicating them to fpl floder. We do the same in case if fpl identity, we use the one from aat folder.

lewisbirks:
@adusumillipraveen thanks for explaining, don't think that really applies to the idam one then as it seems to be configuring a lot of other things instead of vault access

adusumillipraveen (approved):

PR #9307, opened 2021-04-21
switching demo back for trustcorps

elvianihuseyin:
switching demo back for trustcorps

### Change description ###
switching demo back for trustcorps

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/demo/common/probate/probate-back-office.yaml
+++ b/k8s/demo/common/probate/probate-back-office.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: probate
   annotations:
     fluxcd.io/automated: "true"
-    fluxcd.io/tag.java: glob:prod-*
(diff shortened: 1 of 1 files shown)


damongreen123 (approved):

k8s/demo/common/probate/probate-back-office.yaml
```diff
@@ -20,7 +20,7 @@ spec:
     java:
       replicas: 2
       useInterpodAntiAffinity: true
-      image: hmctspublic.azurecr.io/probate/back-office:prod-22222a28
+      image: hmctspublic.azurecr.io/probate/back-office:latest
```
vinceholt, line 23: I think this should be pr-1413-dddaec01 shouldn't it so that we don't get overwritten with master or some other image? Or is this ensured with the pr-1413-* tag in annotations?

PR #9781, opened 2021-05-17
Revert "RPET-991: release new parameter to ACA API call"

yarpo:
Reverts hmcts/cnp-flux-config#9769

We don't want to merge it, but it's opened in case we would have to roll back release.


jakub-kozlowski (approved):

yarpo:
It looks like we don't need it.

PR #9993, opened 2021-06-02
adding vh-admin-web

vialde:
### Change description ###

Adding config to deploy vh-admin-web to our environment

```
[ ] Yes
[x ] No
```

### DIFF
--- a/k8s/aat/common/vh/identity.yaml
+++ b/k8s/aat/common/vh/identity.yaml
@@ -1,0 +1,20 @@
+apiVersion: "aadpodidentity.k8s.io/v1"
+kind: AzureIdentity
+metadata:
+  name: vh
+  namespace: vh
+spec:
+  type: 0
+  resourceID: 
+  clientID: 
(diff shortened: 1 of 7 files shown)


k8s/aat/common/vh/identity.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: "aadpodidentity.k8s.io/v1"
```
timja, line 1: this should go into https://github.com/hmcts/shared-services-flux,

We don't have any docs in shared-services-flux currently but it's the exact same as far as I know

timja (changes requested):

k8s/aat/common/vh/identity.yaml
vialde: just the identity yaml or this whole branch?

timja: everything in this PR

PR #10121, opened 2021-06-17
Add NO_FAULT_DIVORCE to case types

linusnorton:


ernestman28 (commented):
is that the correct case types or typo? NO_FAULT_DIVORCE18

PR #10208, opened 2021-06-25
Add flux system with sandbox config

adusumillipraveen:

```
[ ] Yes
[x] No
```


apps/flux-system/base/image-automation.yaml
```diff
@@ -0,0 +1,24 @@
...
+  namespace: flux-system
+spec:
+  interval: 1m0s
+  sourceRef:
+    kind: GitRepository
+    name: flux-config
+  git:
+    checkout:
+      ref:
+        branch: main
```
thomast1906, line 14: Do we want to reference master until we switch to main?

adusumillipraveen: Good shout, missed that one.

thomast1906 (approved):

PR #10226, opened 2021-06-28
Add admin apps

adusumillipraveen:
Adding kube-slack and traefik with a basic HR migration script

```
[ ] Yes
[x] No
```

### DIFF
--- a/apps/admin/keyvault-flexvol/keyvault-flexvol.yaml
+++ b/apps/admin/keyvault-flexvol/keyvault-flexvol.yaml
@@ -1,0 +1,48 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    app: keyvault-flexvolume
+  name: keyvault-flexvolume
+  namespace: admin
+spec:
+  selector:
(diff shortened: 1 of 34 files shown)


apps/admin/keyvault-flexvol/keyvault-flexvol.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    app: keyvault-flexvolume
```
timja, line 5: is this really still needed?

timja (approved):

apps/admin/keyvault-flexvol/keyvault-flexvol.yaml
adusumillipraveen: yeah, saw a couple of jobs using it in prod. (we forgot about jobs when we chased people)

adusumillipraveen: Its there to add it only to prod

PR #10272, opened 2021-06-30
fix: Update fee keyword

tomaszpowroznik:


samuelzerozero (approved):
Waiting for the confirmation from Fee And Pay that changes have been reverted in AAT and DEMO

PR #10342, opened 2021-07-07
Update kustomization.yaml

iguannago:
Remove the configuration cron job from AAT until we fix the number of idam class made


alexbaeza (approved):
Temporarily disabling the job while we implement caching

hannah38 (approved):

PR #10799, opened 2021-08-03
Script for creating new env directories in flux v2

adusumillipraveen:
will do a follow up for preview

```
[ ] Yes
[x] No
```

### DIFF
--- a/migration/create-environment.sh
+++ b/migration/create-environment.sh
@@ -4,7 +4,7 @@ set -ex
 ENVIRONMENT=$1
 
 #Logging to cft sbox to pull secrets
-az aks get-credentials --resource-group cft-sbox-00-rg --name cft-sbox-00-aks --subscription b72ab7b7-723f-4b18-b6f6-03b0f2c6a1bb --overwrite-existing -a && kubectl config use-context cft-sbox-00-aks-admin
+az aks get-credentials --resource-group cft-sbox-00-rg --name cft-sbox-00-aks --subscription b72ab7b7-723f-4b18-b6f6-03b0f2c6a1bb --overwrite-existing -a
 
 # ----------------------- Cluster files ----------
 
--- a/migration/create-environment.sh
+++ b/migration/create-environment.sh
@@ -1,0 +1,236 @@
+#!/usr/bin/env bash
+set -ex
+
+ENVIRONMENT=$1
+
+# ----------------------- Cluster files ----------
+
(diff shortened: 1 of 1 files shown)


migration/create-environment.sh
```diff
@@ -0,0 +1,236 @@
...
+mkdir -p $CLUSTER_DIRECTORY/base
+
+cp k8s/$ENVIRONMENT/pub-cert.pem  $CLUSTER_DIRECTORY/pub-cert.pem
+
+(
+cat <<EOF
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../../../apps/flux-system/sbox/base
```
timja, line 23: sbox?

migration/create-environment.sh
```diff
@@ -0,0 +1,236 @@
...
+    substitute:
+      ENVIRONMENT: "$ENVIRONMENT"
+      CLUSTER: "01"
+      ENV_MONITOR_CHANNEL: "aks-monitor-$ENVIRONMENT"
+      KEYVAULT_ENVIRONMENT: "$ENVIRONMENT"
+EOF
+) > "$FLUX_SYSTEM_DIRECTORY/01/kustomize.yaml"
+
+
+kubectx cft-sbox-00-aks-admin
```
timja, line 113: maybe should be a variable at the top for the cluster name? and verify that kubectx is installed?

timja (approved):
comments are non blocking

migration/create-environment.sh
adusumillipraveen: change it to just relogin.

migration/create-environment.sh
```diff
@@ -0,0 +1,237 @@
+#!/usr/bin/env bash
+set -ex
+
+ENVIRONMENT=$1
+
+#Logging to cft sbox to pull secrets
+az aks get-credentials --resource-group cft-sbox-00-rg --name cft-sbox-00-aks --subscription b72ab7b7-723f-4b18-b6f6-03b0f2c6a1bb --overwrite-existing -a && kubectl config use-context cft-sbox-00-aks-admin
```
timja, line 7: is the use-context needed? won't it swap you anyway?

adusumillipraveen: never realised, always had it stored that way :)

thomast1906 (approved):

PR #10986, opened 2021-08-12
LAU-56: Add LAU Flux Kustomize pipeline configuration

bradley-hmcts:

https://tools.hmcts.net/jira/browse/LAU-56

### Change description ###

Add LAU Flux pipeline configuration.

Only AAT and Demo are configured initially.

Managed identity configuration is still outstanding but will be added when LAU shared infrastructure is configured.

```
[ ] Yes
[x] No
```


k8s/namespaces/lau/lau-case-backend/lau-case-backend.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: lau-case-backend
+spec:
+  releaseName: lau-case-backend
+  chart:
+    git: git@github.com:hmcts/hmcts-charts
+    ref: master
+    path: stable/lau-case-backend
+  values:
+    java:
+      memoryRequests: "1024Mi"
```
timja, line 13: have you tested that this is necessary?

overriding defaults should only be done if proven via performance testing

k8s/namespaces/lau/lau-case-frontend/lau-case-frontend.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: lau-case-frontend
+spec:
+  releaseName: lau-case-frontend
+  chart:
+    git: git@github.com:hmcts/hmcts-charts
+    ref: master
+    path: stable/lau-case-frontend
+  values:
+    nodejs:
+      memoryRequests: "1024Mi"
```
timja, line 13: is this needed? frontend apps normally don't need to override the defaults unless they are large projects

timja (commented):
can you look at my comments please

k8s/namespaces/lau/lau-case-backend/lau-case-backend.yaml
bradley-hmcts: I've removed these overrides. Copied what we did with PCQ as the apps are of similar size.

k8s/namespaces/lau/lau-case-frontend/lau-case-frontend.yaml
bradley-hmcts: I've removed these overrides. Copied what we did with PCQ as the apps are of similar size.

timja (approved):

PR #11121, opened 2021-08-18
add CAMUNDA_MAX_RESULTS env var to the prod patch to target 1000 task…

iguannago:
set max result to 1000 in Prod because in AAT the default is 10 tasks to reduce idam request calls.


iguannago:
no longer needed

PR #11132, opened 2021-08-19
Test imageautomation on sbox Jenkins

thomast1906:
### Change description ###
Test imageautomation on sbox Jenkins
- Some kustomize is commented out, required as kustomize will fail due to missing initial components - will PR these at a later stage once base/image-automation-components.yaml is deployed

```
[ ] Yes
[X] No
```


timja (approved):
approved but should be reverted after as you're just adding the same config

PR #11428, opened 2021-09-03
clear out the update case job

iguannago:


mkayad:
work done in https://github.com/hmcts/cnp-flux-config/pull/11408

PR #11774, opened 2021-09-28
Add Prod finrem-ns to Kustomize

damongreen123:
https://tools.hmcts.net/jira/browse/DTSSE-510

```
[ ] Yes
[ ] No
```


k8s/prod/common-overlay/financial-remedy/kustomization.yaml
```diff
@@ -3,3 +3,6 @@ kind: Kustomization
 namespace: financial-remedy
 bases:
 - ../../../namespaces/financial-remedy
+- ../../../namespaces/financial-remedy/finrem-ns/finrem-ns.yaml
```
timja, line 6: normally you would add this to the common base at this point rather than adding it to the prod kustomization

timja (approved):

timja (changes requested):
wait don't you need to remove these from all the environment specific additions you already did?

damongreen123:
> wait don't you need to remove these from all the environment specific additions you already did?

Makes sense now. like the latest commit?

timja (approved):
thanks

ruban72 (approved):

PR #11813, opened 2021-09-30
Create ithc env

adusumillipraveen:

```
[ ] Yes
[x] No
```


chrisorisawayi (dismissed):

apps/monitoring/ithc/base/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../base
+  - prometheus-values.yaml
```
hannah38, line 5: Filename seems to be monitoring-values.yaml in this case

adusumillipraveen: good shout, realised its wrong in perftest too. Aligned them

hannah38 (approved):

PR #12162, opened 2021-10-19
Use pre-release flux2 for testing

timja:
See https://github.com/fluxcd/source-controller/issues/439#issuecomment-943580625


apps/monitoring/base/kustomize.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
+t apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
```
timja, line 1:
```suggestion
apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
```

PR #12379, opened 2021-10-28
updated traefik for new demo AKS cluster

darren221980:

https://tools.hmcts.net/jira/browse/DTSPO-5329

### Change description ###

updated traefik to use new AKS cluster

```
[ ] Yes
[ x] No
```


k8s/namespaces/admin/traefik-private/patches/demo-01.yaml
```diff
@@ -5,4 +5,4 @@ metadata:
   namespace: admin
 spec:
   values:
-    loadBalancerIP: 10.51.31.251
\ No newline at end of file
+    loadBalancerIP: 10.50.80.251
```
thomast1906, line 8: Can you review these IPs? they don't look correct compared to last 10.50.XXX.0 of each subnet

k8s/namespaces/admin/traefik-private/patches/demo-00.yaml
```diff
@@ -5,4 +5,4 @@ metadata:
   namespace: admin
 spec:
   values:
-    loadBalancerIP: 10.51.15.221
\ No newline at end of file
+    loadBalancerIP: 10.50.79.221
```
hannah38, line 8: Hi Darren, can you review these - so k8s/ folder is flux v1 and this is on v2 for demo now - but it should be quick to find those v2 files instead

darren221980: undone the previous change for flux v1 and put changes into the flux v2 folder only.

PR #12545, opened 2021-11-08
DTSPO-5506 Gloden path Github app entry

mokainos:


apps/jenkins/jenkins/jenkins.yaml
```diff
@@ -390,6 +390,11 @@ spec:
                           description: "GitHub APP Key - prl"
                           id: "hmcts-jenkins-prl"
                           privateKey: "${hmcts-jenkins-prl-ghapp}"
+                      - gitHubApp:
+                          appID: "149849"
+                          description: "GitHub APP Key - cft golden path"
```
timja, line 395: this is not to do with golden path, it's a general app for use so we don't have to keep creating new ones.
```suggestion
                          description: "GitHub APP Key - cft"
```

timja (approved):

PR #12593, opened 2021-11-09
DTSPO-5506 Golden Path:  Create NS

mokainos:


k8s/namespaces/labs/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
timja, line 1: this needs to be added to the environment you are using it in

thomast1906: @mokainos  you need a folder setup like this 
https://github.com/hmcts/cnp-flux-config/tree/master/apps/rpe/base

Doc on setup
https://github.com/hmcts/cnp-flux-config/blob/master/docs/repo-setup.md
https://github.com/hmcts/cnp-flux-config/blob/master/docs/app-deployment.md

timja (dismissed):

timja (approved):

PR #12842, opened 2021-11-22
Try tune the tcp_retries value

nikola-naydenov-hmcts:


verityit (approved):

apps/idam/idam-web-public/preview.yaml
```diff
@@ -4,6 +4,10 @@ metadata:
   name: idam-web-public
   namespace: idam
 spec:
+  securityContext:
```
timja, line 7: this won't do anything, doubt it's available inside the chart and everything in the chart is under values.java anyway

PR #13045, opened 2021-12-02
Update base kustomize with /base and update apiVersion

thomast1906:
### Change description ###
Update base kustomize with /base and update apiVersion

```
[ ] Yes
[X] No
```


timja (approved):
blind approval

PR #13152, opened 2021-12-08
test only perftest00

thomast1906:
### Change description ###
test only perftest00

```
[ ] Yes
[X] No
```


nickmoyni (commented):
I still think tolerations code might be better positioned direct under values, rather than under cve give it a generic toleration for all.

thomast1906:
closing for now

PR #13161, opened 2021-12-08
CCD-2318: Onboard Reform Employment Tribunal services

rajaram-nagarajan:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CCD-2318

### Change description ###

Enabled services in CCD def-store (Demo and AAT env)

```
[ ] Yes
[x] No
```


hemantt (approved):

MSancaktutar (commented):
It's not very common for Service apps to need to consume Def Store. Would the Service app actually need it? If so, could we kindly ask for what requirement, just for our awareness?

rajaram-nagarajan:
The service does not need to consume Def Store.

PR #13202, opened 2021-12-10
IDAM bump idam-web-public chart

nikola-naydenov-hmcts:


KianKwa:
This has been checked to be the same as the preview branch chart version on the web public project

KianKwa (approved):

PR #13261, opened 2021-12-14
added adoption preview common overlay kustomization

abidfs:
### Change description ###
added adoption preview common overlay kustomization

[ ] Yes
[x] No


timja (changes requested):
Preview uses flux v2, see https://github.com/hmcts/cnp-flux-config/blob/master/apps/adoption/preview/base/kustomization.yaml

PR #13289, opened 2021-12-15
swap key

thomast1906:
### Change description ###
swap key

```
[ ] Yes
[X] No
```


nickmoyni (approved):
LG, would like to understand difference here.

PR #13419, opened 2021-12-24
Ccd 2379

Thor-tech-of-metal:
CCD-2379: CCD Demo Elastic search is throwing 502 error
    
               [CCD-2379] (https://tools.hmcts.net/jira/browse/CCD-2379).


hemantt (changes requested):
Replace them with Prod images?

MSancaktutar (approved):

PR #13567, opened 2022-01-12
CCD-2420: Probate PROD - Allow ad-hoc data extract requests to complete

rajaram-nagarajan:

https://tools.hmcts.net/jira/browse/CCD-2420

### Change description ###

Changed to support probate batch rerun

```
[ ] Yes
[x] No
```


k8s/namespaces/ccd/ccd-data-store-api/prod.yaml
```diff
@@ -29,5 +29,5 @@ spec:
         IDAM_OIDC_URL: https://hmcts-access.service.gov.uk
         LOGGING_LEVEL_UK_GOV_HMCTS_CCD_SECURITY_FILTERS: DEBUG
         REQUEST_SCOPE_CACHED_CASE_TYPES: GrantOfRepresentation
-        REQUEST_SCOPE_CACHED_CASE_TYPES_FROM_HOUR: 1
+        REQUEST_SCOPE_CACHED_CASE_TYPES_FROM_HOUR: 20
```
MSancaktutar, line 32: this doesn't work, unfortunately.
The hour limits are applied within the same day.
I suggest something like 20-24. And then, the revert will bring it back to 1-6.
I can merge the revert PR just before midnight.

MSancaktutar (changes requested):
needs correction

k8s/namespaces/ccd/ccd-data-store-api/prod.yaml
rajaram-nagarajan: You are right, Mutlu. Just checked the condition in the data-store. In that case, yes, we can go for 20-24.

MSancaktutar (dismissed):
lgtm

MSancaktutar (approved):
lgtm

PR #13666, opened 2022-01-19
Patches aad-pod-id on perftest and ithc

ciaranca:

DTSPO-5407

### Change description ###

Patch ITCH and perftest aad-pod-id


nickmoyni (approved):
LG

PR #13675, opened 2022-01-19
Ensure kubernetes job is only run once and not in parallel [Lower ENVS]

alexbaeza:

https://tools.hmcts.net/jira/browse/RWA-1066

### Change description ###

Ensures kubernetes cron jobs are run independently every 2 minutes alternating between clusters.

This PR contains the prod patches but SHOULD NOT APPLY TO PROD.
Prod deployment will be enabled as part of release 2.

```
[ ] Yes
[x ] No
```


timja (dismissed):
not checked carefully

k8s/namespaces/wa/wa-task-batch-configuration/demo-00.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
+  name: wa-task-batch-configuration
+spec:
+  values:
+    job:
+      schedule: "0,4,8,12,16,20,24,28,32,36,40,44,48,52,56 * * * *"
```
raghera, line 8: Not sure we want this one to be more frequent?  The Configuration job is more for R1 when it failed.  I think we only care about R2 in `demo`. 

Shouldn't do any harm though.

vgulapally: done

satvinder-hullait (approved):

dharmendrak (approved):

timja (approved):

PR #13699, opened 2022-01-20
created lab flux config louisehuyton-nodejs

louisehuyton:
DTSPO-5507

### Change description ###
Ran `create-lab-flux-config.sh` script to create flux config required for lab

```
[ ] Yes
[X] No
```

### DIFF
--- a/apps/labs/labs-louisehuyton-nodejs/labs-louisehuyton-nodejs.yaml
+++ b/apps/labs/labs-louisehuyton-nodejs/labs-louisehuyton-nodejs.yaml
@@ -1,0 +1,26 @@
+---
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
+metadata:
+  name: labs-louisehuyton-nodejs
+  namespace: labs
+spec:
+  releaseName: labs-louisehuyton-nodejs
+  chart:
+    spec:
+      chart: ./stable/labs-louisehuyton-nodejs
+      sourceRef:
+        kind: GitRepository
+        name: hmcts-charts
+        namespace: flux-system
+      interval: 1m
+  values:
+    nodejs:
+      image: hmctssandbox.azurecr.io/labs/louisehuyton-nodejs:latest # {"$imagepolicy": "flux-system:labs-louisehuyton-nodejs"}
+      ingressHost: labs-louisehuyton-nodejs-{{ .Values.global.environment }}.platform.hmcts.net
+      disableTraefikTls: true
+      environment:
+        RECIPE_BACKEND_URL: http://labs-louisehuyton-nodejs-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
+    global:
+      environment: sandbox
+      tenantId: "531ff96d-0ae9-462a-8d2d-bec7c0b42082"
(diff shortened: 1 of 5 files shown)


zc-hmcts (approved):

apps/labs/labs-louisehuyton-nodejs/labs-louisehuyton-nodejs.yaml
```diff
@@ -0,0 +1,26 @@
...
+      chart: ./stable/labs-louisehuyton-nodejs
+      sourceRef:
+        kind: GitRepository
+        name: hmcts-charts
+        namespace: flux-system
+      interval: 1m
+  values:
+    nodejs:
+      image: hmctssandbox.azurecr.io/labs/louisehuyton-nodejs:latest # {"$imagepolicy": "flux-system:labs-louisehuyton-nodejs"}
+      ingressHost: labs-louisehuyton-nodejs-{{ .Values.global.environment }}.platform.hmcts.net
```
timja, line 20: this should be 
```suggestion
      ingressHost: labs-louisehuyton-nodejs.{{ .Values.global.environment }}.platform.hmcts.net
```

but it should be inherited from your chart, i.e. it doesn't need to be added here

apps/labs/labs-louisehuyton-nodejs/labs-louisehuyton-nodejs.yaml
```diff
@@ -0,0 +1,26 @@
...
+      disableTraefikTls: true
+      environment:
+        RECIPE_BACKEND_URL: http://labs-louisehuyton-nodejs-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
+    global:
+      environment: sandbox
+      tenantId: "531ff96d-0ae9-462a-8d2d-bec7c0b42082"
```
timja, line 26: this section isn't needed, where did it come from?
```suggestion
```

timja (changes requested):
looks pretty close, would be good to know where some of these things came from, if it's the script then it needs fixing

apps/labs/labs-louisehuyton-nodejs/labs-louisehuyton-nodejs.yaml
louisehuyton: Can see where the '-' is added instead of '.' in the script. Would that need fixed or completely removed from the script due to be being inherited @timja?

timja: should be completely removed ingress hosts should be setup in the apps chart

louisehuyton: cool will do thank you!

apps/labs/labs-louisehuyton-nodejs/labs-louisehuyton-nodejs.yaml
louisehuyton: Found this within the add-helm release script [here](https://github.com/hmcts/cnp-flux-config/blob/021534d777e678d6b08891d7dbdf9f2711efd1de/bin/v2/add-helm-release.sh#L68) which is ran by the create-lab-flux-config script.
Should this be removed also @timja?

PR #13722, opened 2022-01-21
Preview quotas for highest usage teams

hannah38:
Quota is set to the same as cmc has, they had the same issue with many ccd pods also being spun up.
These teams can clear releases that are no longer needed to make space for new ones.

In next sprint will change this to lower duplication, and give all teams a max usage on preview.

```
[ ] Yes
[ x] No
```


cakeben (approved):

jordankainos (approved):
lgtm

timja (approved):

PR #13727, opened 2022-01-21
updated ithc env to use updated prometheus version

darren221980:

https://tools.hmcts.net/jira/browse/DTSPO-5413

### Change description ###

Updated ITHC env to use Prometheus version 30.0.1 (as been tested in sbox already). Note the referenced kube-prometheus-stack file in the patchesStrategicMerge is from the sbox env to save duplicating this. Once confirmed as working in all env's this will be removed as the main kube-prometheus-stack file will be updated (which all envs use) to use version 30.0.1.

```
[ ] Yes
[x] No
```


ciaranca (approved):
LG

PR #14178, opened 2022-02-14
change cron times

mdayican:
### Change description ###

change run time and revert previous clustor setting

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/namespaces/em/em-hrs-ingestor/prod-common.yaml
+++ b/k8s/namespaces/em/em-hrs-ingestor/prod-common.yaml
@@ -9,4 +9,4 @@ spec:
         IDAM_API_BASE_URI: "https://idam-api.platform.hmcts.net"
         OPEN_ID_API_BASE_URI: "https://hmcts-access.service.gov.uk/o"
         ENABLE_CRONJOB: true
-        MAX_FILES_TO_PROCESS: 1000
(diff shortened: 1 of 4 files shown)


apereverzin (dismissed):

apereverzin (approved):

k8s/namespaces/em/em-hrs-ingestor/prod-common.yaml
```diff
@@ -9,4 +9,4 @@ spec:
         IDAM_API_BASE_URI: "https://idam-api.platform.hmcts.net"
         OPEN_ID_API_BASE_URI: "https://hmcts-access.service.gov.uk/o"
         ENABLE_CRONJOB: true
-        MAX_FILES_TO_PROCESS: 2500
+        MAX_FILES_TO_PROCESS: 6000
```
yogesh-hullatti, line 12: Should we just leave this at 2500 only? So that ccd-dat-store-api does not get overloaded while creating case?

mdayican: job will run only once, it will start at 11:30. we can not create that fast. 
and we have limit on our queue 100. after that we ignore the add requests.

yogesh-hullatti (approved):

PR #14260, opened 2022-02-16
Added in new attributes for Demo for Bulk scan orchestrator

yogesh-hullatti:
https://tools.hmcts.net/jira/browse/DTSBPS-839
### Change description ###
Added in new attributes for Demo for Bulk scan orchestrator
```
[ ] Yes
[X] No
```


mdayican (approved):

apereverzin (approved):

mdayican (changes requested):
why do we need this?

yogesh-hullatti:
> why do we need this?

As the new properties/attributes/secrets needs to be made available to the pods via flux. as the existing chart version does not have these new attributes in it. We have referred to new PR image with old/existing Helm Chart version of 0.4.8. New version of chart - 0.5.1 will be published only after PR is merged.

mdayican (approved):

PR #14312, opened 2022-02-21
Revert "DTSPO-6658 Apply Dynatrace CRDs"

chrisorisawayi:
Reverts hmcts/cnp-flux-config#14311

CRDs successfully applied in AAT but failed in Prod due to error below:

`dynatrace-crd    False     CustomResourceDefinition/oneagents.dynatrace.com dry-run failed, error: failed to convert new object: unable to convert unstructured object to apiextensions.k8s.io/v1, Kind=CustomResourceDefinition: cannot convert int64 to float64   16m`

Will review prod error and revisit change later.

### DIFF
--- a/apps/monitoring/base/kustomize.yaml
+++ b/apps/monitoring/base/kustomize.yaml
@@ -6,7 +6,6 @@ metadata:
 spec:
   dependsOn:
     - name: prometheus-crd
-    - name: dynatrace-crd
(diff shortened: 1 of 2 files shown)


timja (approved):
old cluster version?

chrisorisawayi:
> old cluster version?

Yes looks like.

PR #14438, opened 2022-02-25
Look for presence of Flux issue in DEMO

BWeeks-JusticeGovUK:
Use PR-978 PFE Build to attempt to replicate CNP Flux issue from AAT in DEMO


k8s/namespaces/divorce/div-pfe/demo.yaml
```diff
@@ -8,7 +8,7 @@ spec:
     nodejs:
       ingressClass: traefik-no-proxy
       ingressHost: petitioner-frontend-aks.demo.platform.hmcts.net
-      image: hmctspublic.azurecr.io/div/pfe:prod-8d7a138-20220209162752 #{"$imagepolicy": "flux-system:demo-div-pfe"}
+      image: hmctspublic.azurecr.io/div/pfe:pr-978-89a0537-20220225104131 #{"$imagepolicy": "flux-system:div-pfe"}
```
EllisD-B, line 11: I don't think this is correct - if you want to test on demo you still need to refer to the demo image policy

BWeeks-JusticeGovUK: Good catch - thanks

EllisD-B (approved):

PR #14520, opened 2022-03-02
CCD-2552 add regression environment

nathan-clark:
Create 'regression' env in demo only, I assume the enabled components are sufficient. 

https://tools.hmcts.net/jira/browse/CCD-2552

```
[ ] Yes
[x] No
```


k8s/namespaces/ccd/ccd-regr/ccd-regr.yaml
```diff
@@ -0,0 +1,106 @@
...
+      idam-pr: false
+    ccd:
+      apiGatewayWeb:
+        enabled: true
+      emAnnotation:
+        enabled: true
+      draftStore:
+        enabled: true
+      dmStore:
+        enabled: true
```
MSancaktutar, line 22: ccd chart may need to be upgraded for versions of these ones

timja (approved):

k8s/namespaces/ccd/ccd-regr/ccd-regr.yaml
```diff
@@ -0,0 +1,106 @@
...
+        environment:
+          AZURE_STORAGE_DEFINITION_UPLOAD_ENABLED: true
+    ccd-data-store-api:
+      java:
+        image: hmctspublic.azurecr.io/ccd/data-store-api:prod-0b87f89-20220301173815 #{"$imagepolicy": "flux-system:ccd-data-store-api"}
+        environment:
+          CCD_DM_DOMAIN: https?://.*dm-store.*
+          ELASTIC_SEARCH_ENABLED: true
+          ELASTIC_SEARCH_HOSTS: "{{ .Release.Name }}-es-master:9200"
+          ELASTIC_SEARCH_DATA_NODES_HOSTS: "http://{{ .Release.Name }}-es-master:9200"
```
MSancaktutar, line 94: Are these the only env vars that need to be specified here? At least a few more looks to be required to be overridden, as they are in data store demo.

k8s/namespaces/ccd/ccd-regr/ccd-regr.yaml
```diff
@@ -0,0 +1,106 @@
+apiVersion: helm.fluxcd.io/v1
+kind: HelmRelease
+metadata:
+  name: ccd-regr
+spec:
+  releaseName: ccd-regr
```
MSancaktutar, line 6: Where does the releaseName come from? Proposed by SE? I would prefer naming it like `regtest`, `rt` or `str`. Obviously, it's an environment that we don't want to have regressions but to test if there are any. :)

thomast1906 (approved):

MSancaktutar (approved):

PR #14605, opened 2022-03-07
Add startup probe in SBOX

reespozzi:
### Change description ###
Testing startup probe configuration is correct in sandbox first before promoting to prod, checking that the path and port are correctly configured.

```
[ ] Yes
[x] No
```

### DIFF
--- a/apps/jenkins/jenkins/sbox-intsvc/jenkins.yaml
+++ b/apps/jenkins/jenkins/sbox-intsvc/jenkins.yaml
@@ -10,7 +10,7 @@ spec:
       healthProbes: true
       probes:
         startupProbe:
-          failureThreshold: 12
+          failureThreshold: 15
       javaOpts: >-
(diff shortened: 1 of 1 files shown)


apps/jenkins/jenkins/sbox-intsvc/jenkins.yaml
```diff
@@ -7,6 +7,14 @@ spec:
   values:
     checkDeprecation: false
     controller:
+      healthProbes: true
+      probes:
+        startupProbe:
+          httpGet:
+            path: '/login'
+            port: http
```
reespozzi, line 15: Not sure if port needs explicitly declared here

timja: can you try by just overriding what you want to override rather than specifying the defaults

PR #14907, opened 2022-03-22
DTSPO-7372 Add tolerations

mokainos:


thomast1906 (approved):
Once tested, could we move this into the base?

PR #14922, opened 2022-03-22
DTSPO-7372 Add tolerations

mokainos:


thomast1906 (commented):
It this to test? I see @mokainos had something similar earlier

PR #14963, opened 2022-03-24
Dynatrace-upgrade-perftest-01-testing

nickmoyni:

https://tools.hmcts.net/jira/browse/DTSPO-6187

### Change description ###

Adding oneagent classicfullstack code. Testing against perftest-01.

```
[ ] Yes
[x] No
```


apps/monitoring/perftest/01/dynatrace-operator.yaml
```diff
@@ -0,0 +1,40 @@
+---
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
+metadata:
+  name: dynatrace-operator
+  namespace: monitoring
```
chrisorisawayi, line 6: Is this being tested in monitoring namespace first before changing to dynatrace?

apps/monitoring/perftest/01/dynatrace-operator.yaml
```diff
@@ -0,0 +1,40 @@
...
+    spec:
+      chart: dynatrace-operator
+      version: 0.4.2
+      sourceRef:
+        name: dynatrace-operator
+        kind: HelmRepository
+        namespace: monitoring
+  valuesFrom:
+    - kind: Secret
+      name: oneagent-values
```
chrisorisawayi, line 20: Same as base and should not be required.

chrisorisawayi, line 20: Same as base and should not be required.

apps/monitoring/perftest/01/dynatrace-operator.yaml
nickmoyni: Ideally like to test under the current monitoring namespace. I did try to use dynatrace but this kept giving me an error.

apps/monitoring/perftest/01/dynatrace-operator.yaml
nickmoyni: Keeping this code, in the event i test with an alternative namespace.

chrisorisawayi: In that case, the other elements apart from the namespace can be removed as these are already in the base.  Example

```
chart:
  spec:
    sourceRef:
      namespace: monitoring
```

nickmoyni: Reduced the code as suggested.

apps/monitoring/perftest/01/dynatrace-operator.yaml
nickmoyni: I attempted to change namespace and this is where the error occurs.

chrisorisawayi (approved):

PR #15006, opened 2022-03-25
 Switching code to use dynatrace ns.

nickmoyni:

https://tools.hmcts.net/jira/browse/DTSPO-6187

### Change description ###

Adding dynatrace as the ns after testing with original monitoring ns.

```
[ ] Yes
[x] No
```


apps/monitoring/perftest/01/dynatrace-operator.yaml
```diff
@@ -3,29 +3,27 @@ apiVersion: helm.toolkit.fluxcd.io/v2beta1
 kind: HelmRelease
 metadata:
   name: dynatrace-operator
-  namespace: monitoring
+  namespace: dynatrace
```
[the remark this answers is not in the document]
nickmoyni: Created new dynatrace name space.

apps/dynatrace/perftest/01/dynatrace-operator2.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: dynatrace-operator
+  namespace: dynatrace
+spec:
+  values:
+    classicFullStack:
+      args:
+        - --set-host-group=PERF_CFT
+        - --set-network-zone=azure.cft
+        - --set-host-id-source=ip-addresses
```
thomast1906, line 13: This entry` --set-host-id-source=ip-addresses` currently breaks Dynatrace - can you review SDS setup?

More details on issue: https://one.dynatrace.com/hc/en-us/restricted?return_to=https%3A%2F%2Fone.dynatrace.com%2Fhc%2Fen-us%2Frequests%2F34601

apps/dynatrace/perftest/01/dynatrace-operator2.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: HelmRelease
+metadata:
+  name: dynatrace-operator
+  namespace: dynatrace
+spec:
+  values:
+    classicFullStack:
+      args:
+        - --set-host-group=PERF_CFT
+        - --set-network-zone=azure.cft
```
thomast1906, line 12: No need for` --set-network-zone` on base its already added

apps/dynatrace/dynatrace-operator2.yaml
```diff
@@ -0,0 +1,37 @@
...
+  namespace: dynatrace
+spec:
+  releaseName: dynatrace-operator
+  interval: 5m
+  chart:
+    spec:
+      chart: dynatrace-operator
+      version: 0.4.2
+      sourceRef:
+        name: dynatrace-operator2
```
thomast1906, line 14: Do we need the 2 reference(s)?

thomast1906:
Do we need this as part of our setup? 
https://github.com/hmcts/sds-flux-config/blob/master/apps/dynatrace/dynatrace-crds/kustomization.yaml

apps/dynatrace/perftest/01/dynatrace-operator2.yaml
nickmoyni: Fixed.

apps/dynatrace/dynatrace-operator2.yaml
nickmoyni: Using the 2nd reference and the helm chart I'm pulling down this name see helm repo file.

apps/dynatrace/base/kustomize.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Kustomization
+metadata:
+  name: dynatrace
+  namespace: flux-system
+spec:
+  path: ./apps/dynatrace/${ENVIRONMENT}/base
+  postBuild:
+    substitute:
+      NAMESPACE: "dynatrace"
+      TEAM_NOTIFICATION_CHANNEL: "dtspo-yellow"
```
thomast1906, line 11: I dont think it should be `dtspo-yellow` - @hannah38  ?

apps/dynatrace/dynatrace-operator2.yaml
thomast1906: I would prefer to keep all without a 2 incoming helm repo etc

apps/dynatrace/base/kustomize.yaml
nickmoyni: I was going to keep as this during my testing then I can revert back to whatever is recommend.

nickmoyni: platops-build-notices - configured to be this.

apps/dynatrace/perftest/base/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../base
+  - ../../dynatrace-operator2.yaml
```
hannah38, line 5: Can we clean up this naming now that these have been to moved to a separate namespace/folder - by removing that 2

apps/dynatrace/perftest/base/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
...
+  - ../../dynatrace-operator-helmrepo2.yaml
+namespace: dynatrace
```
hannah38, line 7: Similar to a comment below - new one is being added to both clusters from here, but then you're only adding a patch for 01. Can you review please

apps/dynatrace/perftest/base/kustomization.yaml
nickmoyni: Done

apps/dynatrace/perftest/base/kustomization.yaml
nickmoyni: Added patch for 00, I was orig only going to test against 01.

darren221980 (approved):

apps/dynatrace/perftest/00/dynatrace-operator.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
+kind: HelmRelease
+metadata:
+  name: dynatrace-operator
+  namespace: dynatrace
+spec:
+  values:
```
thomast1906, line 8: I would have this as a base for perftest, rather than duplication to both perftest/00perftest /01

thomast1906 (changes requested):
I would have this as a base for perftest, rather than duplication to both perftest/00perftest /01

apps/dynatrace/dynatrace-crds/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
thomast1906, line 2: How are the Dynatrace-crds being applied?

nickmoyni: CRD path added to base.

apps/dynatrace/base/kustomize.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
+kind: Kustomization
+metadata:
+  name: dynatrace
+  namespace: flux-system
+spec:
+  path: ./apps/dynatrace/${ENVIRONMENT}/base
```
timja, line 7: needs to depend on the crd https://github.com/hmcts/sds-flux-config/blob/master/apps/dynatrace/base/kustomize.yaml#L8-L9

and you should have an interval set
```suggestion
  interval: 10m
  path: ./apps/dynatrace/${ENVIRONMENT}/base
  dependsOn:
    - name: dynatrace-crd
```

nickmoyni: Fixed, thanks for pointing this out.

timja (approved):

apps/dynatrace/base/kustomize.yaml
```diff
@@ -0,0 +1,23 @@
...
+  namespace: flux-system
+spec:
+  interval: 10m
+  path: ./apps/dynatrace/${ENVIRONMENT}/base
+  dependsOn:
+    - name: dynatrace-crd
+  postBuild:
+    substitute:
+      NAMESPACE: "dynatrace"
+      TEAM_NOTIFICATION_CHANNEL: "platops-build-notices"
```
thomast1906, line 14: I think on this, could we create an environmenet variable for `platops-build-notices`? Moving forward with Flux v2 we have other apps that will reference this such as plum/camunda etc

thomast1906: @hannah38 @nickmoyni  what you think?

timja: trivial to find replace in the repo 👅

thomast1906 (approved):

hannah38 (approved):

PR #15193, opened 2022-04-05
seal secret with new name value for Dynatrace

thomast1906:
### Change description ###
seal secret with new name value for Dynatrace

```
[ ] Yes
[X] No
```


nickmoyni (approved):
LGTM

PR #15429, opened 2022-04-22
Flux v1 tuning - Extend git.pollInterval

hannah38:
Tested on perftest clusters where flux was previously failing
Have tested multiple values, was still seeing some timeouts and restarts at 2 mins.
https://github.com/fluxcd/flux/blob/v1.24.3/chart/flux/values.yaml#L103

```
[ ] Yes
[x ] No
```


nickmoyni (approved):
LG

PR #15464, opened 2022-04-26
SSCS-10372: update case loader cron job

Jonathan-Weston1:
Due to increase in traffic we need to change the scheduled time for the caseloader cron job to run when system traffic is low.

https://tools.hmcts.net/jira/browse/SSCS-10372

### Change description ###
Due to increase in traffic we need to change the scheduled time for the caseloader cron job to run when system traffic is low.

```
[ ] Yes
[x] No
```

### DIFF
--- a/k8s/prod/cluster-00/sscs/sscs-case-loader.yaml
+++ b/k8s/prod/cluster-00/sscs/sscs-case-loader.yaml
@@ -13,7 +13,7 @@ spec:
     path: stable/sscs-case-loader
   values:
     job:
-      schedule: "0 0/1 * * *"
+      schedule: "0 2/1 * * *"
       useInterpodAntiAffinity: true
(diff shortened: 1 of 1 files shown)


MohammedHMCTS (dismissed):

k8s/prod/cluster-00/sscs/sscs-case-loader.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     path: stable/sscs-case-loader
   values:
     job:
-      schedule: "0 0/1 * * *"
+      schedule: "0 2/1 * * *"
```
ss-evoco, line 16: “At minute 0 past every hour from 2 through 23.”  Not too sure this is what we are aiming for. We need to put a cap on the retries so it stops by start of business.

ss-evoco (changes requested):

ss-evoco (approved):

PR #15556, opened 2022-05-04
patch to test keda v2 only on demo

Jenniferstrej:


timja (approved):

k8s/namespaces/fees-pay/ccpay-callback-function/ccpay-callback-function.yaml
```diff
@@ -5,9 +5,9 @@ metadata:
 spec:
   releaseName: ccpay-callback-function
   chart:
-    repository: https://hmctspublic.azurecr.io/helm/v1/repo/
-    name: ccpay-callback-function
-    version: 0.2.1
+    git: git@github.com:hmcts/hmcts-charts
```
Jenniferstrej, line 8: Had to change the main file because the chart refs are github based and have different keys.
That's required to make the patch on demo (below) work.

PR #15827, opened 2022-05-19
keda 2 upgrade on prod

Jenniferstrej:
### Change description ###

Keda upgrade to v2 on preview (To be merged after [change request](https://mojcppprod.service-now.com/nav_to.do?uri=change_request.do?sys_id=f758a8f51be78558291765b9bd4bcb62) is approved as we will do all path to live tickets at the same time)

```
[ x] Yes - all applications that use keda v1 need to change to v2 
[ ] No
```

### DIFF
--- a/k8s/namespaces/fees-pay/ccpay-callback-function/prod.yaml
+++ b/k8s/namespaces/fees-pay/ccpay-callback-function/prod.yaml
@@ -3,6 +3,8 @@ kind: HelmRelease
 metadata:
   name: ccpay-callback-function
 spec:
+  chart:
+    ref: keda-v2-test
   values:
     function:
       aadIdentityName: ccpay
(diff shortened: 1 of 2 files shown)


thomast1906:
Should the chart update be added here?

```
spec:
  chart:
    ref: keda-v2-test
```

k8s/namespaces/fees-pay/ccpay-callback-function/prod.yaml
```diff
@@ -3,6 +3,8 @@ kind: HelmRelease
 metadata:
   name: ccpay-callback-function
 spec:
+  chart:
```
thomast1906, line 6: For prod, can we keep this in its base file and remove the patch references?

Jenniferstrej:
I removed the point to a keda 2 file since it's already on the base. This PR just pins the chart to the same version other environments are pinned to. At least until I investigate why  this PR https://github.com/hmcts/ccpay-functions-node/pull/81/files hasn't had an effect on production yet.

cakeben (approved):

PR #16783, opened 2022-07-19
Cft preview traefik v2

reespozzi:
https://tools.hmcts.net/jira/browse/DTSPO-8328

Creating config for Traefik V2 migration in preview, following the SBOX config we tested in 00 cluster. We need internal wildcard cert and platform wildcard cert as usual. 

Will need another internal wildcard cert creating in similar vein to sbox before continuing. Cert should be named `wildcard-service-core-compute-internal`, this can be different if changed in code too, and have subject as `CN=.*service.core-compute-preview.internal`

We could maybe do this just changing 00 before we rebuild the cluster so 01 stays the way it is until we have preview-00 working as expected. This is just an initial PR to get the needed code changes down somewhere.

```
[ ] Yes
[ ] No
```


reespozzi:
Marking as ready to review, please don't merge yet

apps/admin/preview/base/kustomization.yaml
```diff
@@ -3,13 +3,10 @@ kind: Kustomization
 resources:
   - ../../base
   - kube-slack-values.yaml
-  - ../../csi-secret-store-provider-v1.0.1
-  - traefik-values.yaml
```
Jenniferstrej, line 7: `traefik-values.yaml` is overriding only traefik v1 configuration apparently, so both `../../traefik/traefik.yaml` and `traefik-values.yaml` are being moved to cluster 01 config exclusively. That way it should be easier to remove traefik v1 config completely when moving all new config to (and removing old config from) base.

darren221980 (approved):

PR #16787, opened 2022-07-19
DFR-1174 deploy to demo

gmikerlemos:
https://tools.hmcts.net/jira/browse/DFR-1174?devStatusDetailDialog=pullrequest


ruban72 (changes requested):
I think you have to change the demo.yaml not the div-cos.yaml here

PR #16858, opened 2022-07-22
RST-4448 Add c100 Demo

WillTaylor22:


apps/c100/base/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../../base
+  - ../identity/identity.yaml
+  - ../identity/identity.yaml
```
endakelly, line 6: I'd remove this duplicate line

PR #16886, opened 2022-07-25
Remove RPTS from prod

reespozzi:
Not live yet and app not got all relative permissions to upgrade java version, removing from prod so we can do AKS upgrades. 

```
[ ] Yes
[ ] No
```


hannah38:
Hey Rees, still being included from here https://github.com/hmcts/cnp-flux-config/blob/master/k8s/prod/common-overlay/kustomization.yaml#L36 if you can removed. Thanks

hannah38 (approved):

PR #16910, opened 2022-07-26
RWA-1612: Job to unblock UNPROCESSABLE message

rasikasri:

https://tools.hmcts.net/jira/browse/RWA-1612

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/apps/wa/automation/kustomization.yaml
+++ b/apps/wa/automation/kustomization.yaml
@@ -26,6 +26,8 @@ resources:
   - ../wa-workflow-api/image-policy.yaml
   - ../wa-messages-find-problem-messages/image-repo.yaml
   - ../wa-messages-find-problem-messages/image-policy.yaml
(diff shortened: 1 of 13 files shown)


adusumillipraveen (commented):
I am not sure if this has been tested/ OATed, looks risky to go into prod directly. cc @cakeben

rasikasri:
> 

wa-messages-find-problem-messages is already running in aat for quite some time. Are you suggesting to add wa-messages-reset-problem-messages only to aat first and then prod? These jobs only sending an request to wa-case-event-handler. reset message request would not do anything in wa-case-event-hanlder as there are no any message id set up to reset. That's why I put it into prod as well. But, if needed, I can create a separete pr for prod.

timja (approved):

raghera (approved):

PR #17120, opened 2022-08-05
scaled up ccd data store on prod00 for upgrade on prod01

darren221980:

https://tools.hmcts.net/jira/browse/DTSPO-8034

### Change description ###

Scaled up ccd-data-store while on prod 01 while prod 00 is taken offline for aks upgrade.

```
[ ] Yes
[x ] No
```


timja (dismissed):
could we add a patch just overriding the replicas and add the base to each cluster? would be cleaner especially if this is around for a bit, there's risk that someone tries updating a config value while this is being upgraded with this approach

timja (commented):
not quite what i meant, you should be able to reference the prod patch first and then in the one you are overriding just override the replica

timja (approved):
you could simplify this by just deleting the resource in the prod base but this is fine

https://github.com/kubernetes-sigs/kustomize/issues/1593
https://stackoverflow.com/questions/64899665/exclude-resource-in-kustomization-yaml/66074466#66074466

PR #17284, opened 2022-08-15
RWA-1575-set frequency of task clean up job every 30 mins

serdaroguzhanakin:
https://tools.hmcts.net/jira/browse/RWA-1575

### Change description ###

```
[ ] Yes
[X] No
```


k8s/namespaces/wa/wa-task-batch-clean-up/aat-01.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   values:
     job:
-      schedule: "5-59/10 * * * *"
+      schedule: "30 * * * *"
```
raghera, line 8: Will these schedules clash at say 12:00 or 13:00?
Better if we split them?

serdaroguzhanakin: It works like below
next at 2022-08-17 12:30:00
then at 2022-08-17 13:30:00
then at 2022-08-17 14:30:00
then at 2022-08-17 15:30:00
then at 2022-08-17 16:30:00

raghera (approved):

madjava (approved):

PR #17667, opened 2022-09-06
fix demo cos image

rishikrsharma:
fix demo cos image


apps/financial-remedy/finrem-cos/demo-image-policy.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
     hmcts.github.com/prod-automated: disabled
```
KleoG, line 6: you can change this to enabled

KleoG (approved):

apps/financial-remedy/finrem-cos/demo-image-policy.yaml
rishikrsharma: made change

PR #17793, opened 2022-09-13
1.) Remove Flux v1 ET config completely.

petermoores-version1:
https://tools.hmcts.net/jira/browse/RET-2507

### Change description ###
1.) Remove Flux v1 ET config completely. All ET applications already migrated to Flux v2. IMPORTANT: PlatOps please check this is the case before approving?
2.) Update Flux v2 'et-cos' application in demo to use non-Prod image.

IMPORTANT: PlatOps by removing all Flux v1 config for ET, assume this won't affect live services if already migrated to Flux v2 in non-reform deployments?

```
[ ] Yes
[ x] No
```


adusumillipraveen (approved):
You have a conflict. LGTM otherwise, no it shouldn't cause issue in prod as GC is disabled.

hsjhita (approved):

PR #17849, opened 2022-09-15
Added in config for em-ped-api in v2

yogesh-hullatti:
https://tools.hmcts.net/jira/browse/EM-4597
### Change description ###
Added in config for em-ped-api in v2. Did not had to add namespace to aat in v2 as it already exists.
```
[ ] Yes
[X] No
```

### DIFF
--- a/apps/em/automation/kustomization.yaml
+++ b/apps/em/automation/kustomization.yaml
@@ -26,3 +26,7 @@ resources:
   - ../em-anno/image-repo.yaml
   - ../em-anno/demo-image-policy.yaml
   - ../em-anno/image-policy.yaml
+  - ../em-ped-api/image-repo.yaml
+  - ../em-ped-api/image-policy.yaml
+  - ../em-ped-api/image-repo.yaml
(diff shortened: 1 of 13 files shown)


apps/em/automation/kustomization.yaml
```diff
@@ -26,3 +26,7 @@ resources:
   - ../em-anno/image-repo.yaml
   - ../em-anno/demo-image-policy.yaml
   - ../em-anno/image-policy.yaml
+  - ../em-ped-api/image-repo.yaml
+  - ../em-ped-api/image-policy.yaml
+  - ../em-ped-api/image-repo.yaml
+  - ../em-ped-api/image-policy.yaml
```
adusumillipraveen, line 32: duplicate
```suggestion
```

adusumillipraveen (commented):
we still need identities to work, it might break if not

yogesh-hullatti:
Cannot go to prod directly.Needs to go through an OAT before it goes to prod. So, need to start again this time only for AAT

PR #17966, opened 2022-09-22
Increase CPU requests too

adusumillipraveen:

```
[ ] Yes
[x] No
```


timja (approved):
I feel like we could go higher on the memory and probably CPU, but not sure if you've looked at consumption in AAT to see where it fits best

adusumillipraveen:
> I feel like we could go higher on the memory and probably CPU, but not sure if you've looked at consumption in AAT to see where it fits best

This is v1 , running only on prod atm. I have set them based on avg they are using currently.

PR #18591, opened 2022-10-20
Add doc and script for adding MI access from another project

chris-pateman:
### Change description ###
Adding documentation and script on how to give your project access to another projects resources via their Managed Identity

A few developers have questions how to configure this on flux and I find this the best place.
the example is/was the civil teams civil-service is connecting to the aac vault aac-aat. For their pod to have access it needs to use the Managed Identity aac-aat-mi. 
This documentation will enable developers/engineers to config on their own without assistance.

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/bin/v2/add-identity-from-another-project.sh
+++ b/bin/v2/add-identity-from-another-project.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -ex
+declare -A SUBMAP
+
+SUBMAP[aat]="DCD-CNP-DEV"
+SUBMAP[perftest]="DCD-CNP-QA"
+SUBMAP[demo]="DCD-CNP-DEV"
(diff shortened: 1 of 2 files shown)


timja (commented):
could you explain the need a bit more? the description in the PR is about the what and not the why

chris-pateman:
> could you explain the need a bit more? the description in the PR is about the what and not the why

Updated. is this ok?

timja:
Could you be more explicit on teams / what or why they are doing this for?

chris-pateman:
> Could you be more explicit on teams / what or why they are doing this for?

updated and added

timja:
What secret / why?

generally teams aren't supposed to access other vaults at runtime, we want runtime apps to be quite limited. but are they deploying another pod in their namespace for some reason?

PR #18647, opened 2022-10-24
Update dg-docassembly.yaml

damongreen123:
Prod out of sync. Trigger PR
https://tools.hmcts.net/jira/browse/DTSSE-2379


mdayican (changes requested):
this will trigger deployment in all environments. you just need prod.

PR #18652, opened 2022-10-24
Adds aadpodidentity label

ciaranca:
### Change description ###

Adds aadpodidentity label

```
[ ] Yes
[X] No
```

### DIFF
--- a/apps/azureserviceoperator-system/aso/aso.yaml
+++ b/apps/azureserviceoperator-system/aso/aso.yaml
@@ -16,3 +16,5 @@ spec:
   values:
     installCRDs: true
     createAzureOperatorSecret: false
+    podLabels:
+        aadpodidbinding: aso-pod-identity-mi


apps/azureserviceoperator-system/aso/aso.yaml
```diff
@@ -15,4 +15,6 @@ spec:
         namespace: azureserviceoperator-system
   values:
     installCRDs: true
-    createAzureOperatorSecret: false
\ No newline at end of file
+    createAzureOperatorSecret: false
+    podLabels:
+        aadpodidbinding: aso-pod-identity-mi
```
timja, line 20: what's lead you to adding this? I wouldn't have expected this label for workload identity

ciaranca: I had messaged Hannah to say that the workload identity piece was holding the ticket back from progressing, if we wanted to create a separate ticket for workload identity and use aad-pod-identity to progress the the ASO ticket, so I was just going to test that. However I have just spoken with Hannah and she is happy to continue to look at workload identity so going to close this PR off

PR #18728, opened 2022-10-27
No ingressclass being pickedup

reespozzi:
ingressClass isn't being picked up properly, seeing if this makes a difference

```
[ ] Yes
[ ] No
```


thomast1906 (approved):
Approved to test

PR #18740, opened 2022-10-27
Enable private ingressclass, update traefik ingressclass name

reespozzi:
**enable flag added so traefik-private shows as an ingress class**

**moving additional arguments as name overwrite for traefik ingress class not working, currently showing as traefik-auth-proxy which we don't have set**

```
[ ] Yes
[ ] No
```


thomast1906 (approved):
If this works, we need to understand why base does not work?

PR #18795, opened 2022-10-31
Update perftest cron frequency

raghera:
Increase initiation schedule to every 2 mins per cluster resulting in a job running every minute.


serdaroguzhanakin (approved):
LGTM

PR #18927, opened 2022-11-04
Added cron config for bulk list create for demo 01 cluster

nitinprabhu:


sunilnatikar (approved):

k8s/demo/cluster-01-overlay/nfdiv/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
timja, line 1: this needs to be referenced somewhere

nitinprabhu: I am not sure where we need to reference  `k8s/demo/cluster-01-overlay/nfdiv/kustomization.yaml` could you please point me to the config ?

timja: https://github.com/hmcts/cnp-flux-config/blob/master/k8s/demo/cluster-01-overlay/kustomization.yaml

nitinprabhu: thanks @timja

timja (approved):

PR #19696, opened 2022-12-07
RWA-2081[AAT] Add new config for new message job of setting processed state

DennisHoV1:
https://tools.hmcts.net/jira/browse/RWA-2081

### Change description ###
-Add new config for new message job of setting processed state in AAT environment 

```
[ ] Yes
[X] No
```


dharmendrak (approved):

ozzyib (dismissed):

serdaroguzhanakin (dismissed):
LGTM

apps/wa/wa-messages-set-processed-state-messages/aat/01.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: HelmRelease
+metadata:
+  name: wa-messages-set-processed-state-messages
+spec:
+  values:
+    job:
+      schedule: "30 * * * *"
+      environment:
+        JOB_NAME: SET_STATE_TO_PROCESSED_ON_MESSAGES
+        RE_RUN: true
```
raghera, line 11: what does `RE_RUN` do?  It's false in cluster 00 and true here.  Is that correct?

raghera (changes requested):
I think you need to answer the questions in this review before merging.

apps/wa/wa-case-event-handler/aat.yaml
```diff
@@ -10,4 +10,5 @@ spec:
         AZURE_SERVICE_BUS_FEATURE_TOGGLE: true
         TEMP: false
         PROBLEM_MESSAGE_RESET_ID_LIST: 489772
-        PROBLEM_MESSAGE_RESET_NULL_EVENT_TIMESTAMP_ID_LIST: 188590,371679
+        PROBLEM_MESSAGE_RESET_NULL_EVENT_TIMESTAMP_ID_LIST:
+        PROBLEM_MESSAGE_SET_STATE_TO_PROCESSED_ID_LIST: 188590,371679
```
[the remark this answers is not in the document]
raghera: Ok that should be ok as long as they're in the state you expect.

apps/wa/wa-messages-set-processed-state-messages/aat/01.yaml
raghera: I don't know off the top of my head.  Please take a look at the code to see what this env var is used for so we know why we are adding it.  We ought to avoid copy / paste unless we know what it is doing.

DennisHoV1: As discussion with Rasika and Ravi, this dummy env variable (RE_RUN) is used to forcibly restart the pod, it means that the new message job is necessary to keep it, we will remove this variable from aat/00.yaml and aat/01.yaml

ozzyib (approved):

PR #19726, opened 2022-12-08
Migrate nfdiv-cron-alert-application-not-reviewed from Flux V1 to V2 for non prod

nitinprabhu:
https://tools.hmcts.net/jira/browse/NFDIV-2708


apps/nfdiv/nfdiv-cron-alert-application-not-reviewed/demo/00.yaml
```diff
@@ -0,0 +1,28 @@
...
+        NOTIFY_TEMPLATE_APPLICANT2_SIGN_IN_DISSOLUTION_URL: https://nfdiv-end-civil-partnership.demo.platform.hmcts.net/applicant2
+        S2S_URL: http://rpe-service-auth-provider-demo.service.core-compute-demo.internal
+        FEE_API_URL: http://fees-register-api-demo.service.core-compute-demo.internal
+        IDAM_API_REDIRECT_URL: https://div-pfe-demo.service.core-compute-demo.internal/authenticated
+        IDAM_API_BASEURL: https://idam-api.demo.platform.hmcts.net
+        CASE_DATA_STORE_BASEURL: http://ccd-data-store-api-demo.service.core-compute-demo.internal
+        DOC_ASSEMBLY_URL: http://dg-docassembly-demo.service.core-compute-demo.internal
+        DOCUMENT_MANAGEMENT_URL: http://dm-store-demo.service.core-compute-demo.internal
+        PRD_API_BASEURL: http://rd-professional-api-demo.service.core-compute-demo.internal
+        SEND_LETTER_SERVICE_BASEURL: http://rpe-send-letter-service-demo.service.core-compute-demo.internal
```
amoriartyMOJ, line 23: Commenting more for confirmation that, we overwrite env variables as default is AAT variables (which is why we don't overwrite AAT files). If this is true, where are the original default values? Where can I find them. Might be being a bit blind, sorry if I am!

amoriartyMOJ (approved):

sunilnatikar (approved):

apps/nfdiv/nfdiv-cron-alert-application-not-reviewed/demo/00.yaml
nitinprabhu: They fallback to default as defined in our app yaml https://github.com/hmcts/nfdiv-case-api/blob/master/src/main/resources/application.yaml#L440

cakeben (approved):

PR #19892, opened 2022-12-14
Migrate sscs to v2

adusumillipraveen:

```
[ ] Yes
[x] No
```


apps/sscs/aat/01/kustomization.yaml
```diff
@@ -2,4 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - ../base
+  - ../../sscs-case-loader/sscs-case-loader.yaml
```
thomast1906, line 5: Do we not want to include this in aat base rather than 00 & 01?

adusumillipraveen: scripting is easy this way :D Problem is not all apps have them on both clusters.

adusumillipraveen: I can try and find a way, if u think it will be cleaner.

apps/sscs/sscs-bulk-scan/prod.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: helm.fluxcd.io/v1
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
 kind: HelmRelease
 metadata:
   name: sscs-bulk-scan
```
chris-pateman, line 4: If these have no parameters in them, do we need to include them?

adusumillipraveen: Not needed, but not easy to make the script aware of this and remove patches.Its just lift and shift of existing config, minimum needed for migration.

apps/sscs/aat/01/kustomization.yaml
thomast1906: Cheers! Sorry just saw the duplicate of that base reference^

thomast1906 (approved):
QQ - do we want to move `../../sscs-hearings-api/sscs-hearings-api.yaml `& `sscs-bulk-scan.yam`l to base to avoid duplication ?

PR #19936, opened 2022-12-16
Migrate fees-pay demo -int apps to flux v2

adusumillipraveen:
Splitting the changes  as there are too many files.

```
[ ] Yes
[x] No
```


adusumillipraveen:
> Some jobs only on 01 and not 00 too?

missed this, yes. Seem to be the case with all envs.

PR #20330, opened 2023-01-13
DTSPO-11878 Switch ptlsbox pact broker to new flex server db

timja:


apps/pact-broker/pact-broker/sbox-intsvc.yaml
```diff
@@ -4,8 +4,10 @@ metadata:
   name: pact-broker
   namespace: pact-broker
 spec:
+  chart:
+    spec:
+      version: 0.4.0
```
timja, line 9: Requires https://github.com/hmcts/chart-pact-broker/pull/33 to be released before merge of this

msl8r (approved):

PR #20403, opened 2023-01-18
RWA-2172 Handle unprocessable message regarding private law on 18 JAN

DennisHoV1:
https://tools.hmcts.net/jira/browse/RWA-2172

### Change description ###
-There are 5 new unprocessable messages on 18 JAN, all message are related to private law after investigation and these message will be set to "PROCESSED" state

```
[ ] Yes
[X] No
```


mrmsali (approved):
Looks good to me.

PR #20679, opened 2023-02-02
Reduce initiation job frequency

dharmendrak:


serdaroguzhanakin (approved):

apps/wa/wa-task-batch-initiation/aat-01.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   values:
     job:
-      schedule: "1-59/2 * * * *"
+      schedule: "0,20 * * * *"
```
mkayad, line 8: the  2 job could catch up because there is only 1 min difference, would be better to 10 and 30

PR #20968, opened 2023-02-15
DTSPO-12428 Perftest:  migrate sealedsecret to sops

mokainos:


thomast1906:
Will this cause any downtime?

mokainos:
> 
It shouldn't cause any  any downtime. We are just recreating the secret. All the prerequisites are already covered as part of https://github.com/hmcts/cnp-flux-config/pull/20959/files

PR #21258, opened 2023-02-28
Disable external-dns in CFT Demo

reespozzi:
https://tools.hmcts.net/jira/browse/DTSPO-12611
First step in migrating CFT Demo to active/active, to allow us to rebuild the second cluster and begin managing DNS records through other components.

This change:
- Disables External-DNS on CFT demo, to stop DNS records being managed by the external dns app on the clusters.

Similar change done in SDS: https://github.com/hmcts/sds-flux-config/pull/2539

```
[ ] Yes
[ ] No
```

### DIFF
--- a/apps/admin/demo/00/kustomization.yaml
+++ b/apps/admin/demo/00/kustomization.yaml
@@ -4,8 +4,6 @@ resources:
   - ../base
 
 patchesStrategicMerge:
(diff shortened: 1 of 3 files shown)


thomast1906 (approved):
@reespozzi is to confirm / announce downtime etc prior

PR #21284, opened 2023-03-01
DTSPO-10566 - update kube-prometheus-stack sbox

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-10566

### Change description ###
Add updated kube-prometheus-stack crds
Update kps in sbox initially
Base will be updated if all is working

```
[ ] Yes
[x] No
```


apps/monitoring/kube-prometheus-stack/kube-prometheus-stack-upgrade.yaml
```diff
@@ -0,0 +1,185 @@
+apiVersion: helm.toolkit.fluxcd.io/v2beta1
```
timja, line 1: why do you need this extra file? can't you just patch the chart version update in?

endakelly: I was following an example done for Traefik recently. I haven't found any changelogs yet to be able to determine if updating the crds would cause an issue so I was trying to update the stack in sbox only before moving onto the other envs just to be safe. Do you reckon it's fine to update the crds across the board and then just update the chart in sbox?

timja: what you've done for crds is fine, I'm just confused about the stack upgrade itself on why you've added this whole file rather than just a patch to update the chart version

endakelly: I was following something Praveen had done for traefik and he had done it that way. I've just updated with a patch now.

timja (approved):

PR #21808, opened 2023-03-24
template ids added for aat, demo and ithc envs

btensay:
https://tools.hmcts.net/jira/browse/RET-3168

The changes in this pr point to updated or new templates. Hence, the new/updated templates have additional personalisation field. To resolve correctly, this pr presumes having the code changes in et-sya-api (https://github.com/hmcts/et-sya-api/pull/305) deployed. 

```
[X] Yes
[ ] No
```


hsjhita:
@btensay I think we should only set the templates in the other environments once the code has been merged in

hsjhita:
@btensay refer to my previous comment. the template can be perftest but should only be in the other environments when merged in otherwise it will fail on those environments

JustMehmet (dismissed):

apps/et/et-sya-api/ithc.yaml
```diff
@@ -11,4 +11,4 @@ spec:
       environment:
         SUBMIT_CASE_EMAIL_TEMPLATE_ID: 10c31d51-c518-4d73-b2c3-1c02771ef41e
         CY_SUBMIT_CASE_EMAIL_TEMPLATE_ID: f61f7f2a-4825-491b-bc48-0299f1a8ba23
-        SUBMIT_CASE_DOC_UPLOAD_ERROR_EMAIL_TEMPLATE_ID: 3007a1e9-13b0-4bf9-9753-398ea91b8564
+        SUBMIT_CASE_DOC_UPLOAD_ERROR_EMAIL_TEMPLATE_ID: 3007a1e9-13b0-4bf9-9753-398ea91b8564
```
hsjhita, line 14: also don't need this but can leave in

hsjhita (approved):

apps/et/et-sya-api/ithc.yaml
btensay: I agree. I thought it would not matter.

PR #21965, opened 2023-03-29
Remove demo flags

reespozzi:
Removes demo flags from xui apps to allow migration to new setup

```
[ ] Yes
[ ] No
```


apps/xui/demo/base/register-org-ingress.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
   name: register-org-ingress
   namespace: xui
 spec:
+  ingressClassName: traefik
```
reespozzi, line 12: This isn't managed by a chart, so assigning this in code

PR #21996, opened 2023-03-30
Clean up traefik and other demo resources

reespozzi:
https://tools.hmcts.net/jira/browse/DTSPO-12618

This change:
- Removes oauth-proxy
- Removes reply-url-operator
- Removes external-dns in demo, and external-dns-public generally as we only use external-dns-private in preview
- Removes traefik-no-proxy
- Removes traefik-auth-proxy
- Removes external IP patches and renames traefik-private to generic traefik ingress class

Similar change done in SDS here https://github.com/hmcts/sds-flux-config/pull/2555

```
[ ] Yes
[ ] No
```


thomast1906 (approved):

reespozzi:
Will raise separate PR to be merged which removes all `traefik-private` or `traefik-no-proxy` references - will have no effect without external-dns anyway, but saves a 150file+ PR here..

PR #22197, opened 2023-04-13
Revert "Revert "Create generic https->https middlewar redirect, test with AAT-00 Neuvector""

danielwilsonkainos:
Reverts hmcts/cnp-flux-config#22196


danielwilsonkainos:
This introduced error in neuvector helm release but cant figure out why that would be the case after reverting. Now going to revert the revert to re-introduce the error for more investigation

reespozzi (approved):
Approving for testing - please revert again if the same error occurs

PR #22268, opened 2023-04-17
DTSPO-12511 - Update safe settings values

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-12511

### Change description ###
Update safe settings values

```
[ ] Yes
[x] No
```


apps/safe-settings/safe-settings/safe-settings.yaml
```diff
@@ -17,12 +17,21 @@ spec:
...
+        valueFrom:
+          secretKeyRef:
+            name: app-env
+            value: ADMIN_REPO
       APP_ID:
         valueFrom:
           secretKeyRef:
             name: app-env
             value: APP_ID
-      CRON: "0 */1 * * *"
```
timja, line 25: why are most of these values secret? a cron schedule and app id aren't sensitive and it just makes it harder to see what's going on here.

endakelly: It doesn't seem to like mixing and matching secret values with non-secret ones

timja: i guess worth a try but it shouldn't matter

timja (approved):

PR #22323, opened 2023-04-19
Add workload identity resources in SBOX

reespozzi:
This change:
- Adds resources that workload identity needs besides service account added in separate PR.
- Done in generic way but just added to sbox base for now instead of base.
- Patches issuer url per env in sbox

```
[ ] Yes
[ ] No
```


clusters/sbox/base/kustomization.yaml
```diff
@@ -14,6 +14,9 @@ resources:
   - ../../../apps/labs/base/kustomize.yaml
   - ../../../apps/neuvector/base/kustomize.yaml
   - ../../../apps/azureserviceoperator-system/base/kustomize.yaml
+  - workload-identity-federated-credential.yaml
+  - workload-identity-rg.yaml
+  - workload-identity-ua-identity.yaml
```
timja, line 19: this isn't namespace specific so i don't think it will work

reespozzi:
@timja @adusumillipraveen made those suggested changes here - struggling to test with the post-build variables though

timja (approved):

PR #22393, opened 2023-04-24
DTSPO-12662 Update sbox environment for WI

reespozzi:
Currently erroring on not being able to find `managed-identities-sbox-rg` or `cnp-sbox-mi` in the DCD-CFT-Sandbox subscription, as `ENVIRONMENT` is set to `sbox`, which works for the other subscription.

```
[ ] Yes
[ ] No
```


apps/base/kustomize.yaml
```diff
@@ -41,4 +41,5 @@ spec:
       CLUSTER_PATH: "${ENVIRONMENT}/${CLUSTER}"
       CLUSTER_FULL_NAME: "${ENVIRONMENT}-${CLUSTER}"
       ISSUER_URL: "${ISSUER_URL:-not-set-yet}"
+      WI_ENVIRONMENT: "${WI_ENVIRONMENT:-$ENVIRONMENT}"
```
reespozzi, line 44: This should default to `ENVIRONMENT` if `WI_ENVIRONMENT` is not set, which lets us override how workload identity is working without breaking any folder structure being used

hannah38 (approved):

PR #22407, opened 2023-04-24
CFT PTLSBOX - Jenkins: explicitly set serviceaccount, disable creation by chart

danielwilsonkainos:
### Change description ###

- disable creation of new service account through chart for jenkins namespace
- explicitly set to new base serviceaccount for namespace instead

```
[ ] Yes
[x] No
```

### DIFF
--- a/apps/jenkins/jenkins/sbox-intsvc/jenkins.yaml
+++ b/apps/jenkins/jenkins/sbox-intsvc/jenkins.yaml
@@ -6,6 +6,9 @@ metadata:
 spec:
   values:
     checkDeprecation: false
+    serviceAccount:
+      create: false
+      name: jenkins
     controller:
       tag: 2.400-387
       javaOpts: >-


apps/jenkins/jenkins/sbox-intsvc/jenkins.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
 spec:
   values:
     checkDeprecation: false
+    serviceAccount:
+      create: false
+      name: jenkins
```
timja, line 11: I don't think you need this it should just work but ok anyway

danielwilsonkainos: this is the only thing added in the PR - if we remove this then we're using the current config that is what is currently seeing the issue?

danielwilsonkainos: will give it a whirl

timja: it's the `create: false` you need, i don't think you need the name.

PR #22428, opened 2023-04-25
DTSPO-12662 Workload Identity Resources are not tagged

reespozzi:
Pods are giving errors that UA resource does not fit our tagging policy. Not sure how/why this isn't showing for resource group either. Will try and test after this if tags defined in flux are merged with existing azure tags, or if existing azure tags are overridden.

```
[ ] Yes
[ ] No
```

### DIFF
--- a/apps/base/workload-identity/workload-identity-ua-identity.yaml
+++ b/apps/base/workload-identity/workload-identity-ua-identity.yaml
@@ -6,6 +6,11 @@ metadata:
   annotations:
     serviceoperator.azure.com/reconcile-policy: detach-on-delete
 spec:
+  tags:
+    builtFrom: https://github.com/hmcts/cnp-flux-config
+    businessArea: CFT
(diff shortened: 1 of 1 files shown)


apps/base/workload-identity/workload-identity-ua-identity.yaml
```diff
@@ -6,6 +6,11 @@ metadata:
   annotations:
     serviceoperator.azure.com/reconcile-policy: detach-on-delete
```
timja, line 7: what if we set this to skip?

reespozzi: Hm, potentially.. worth a shot as opposed to tagging all app resources as core i guess - and seems to have same behaviour that deletion in k8s doesn't propagate to Azure, will update

hannah38 (approved):

timja (approved):

PR #22444, opened 2023-04-26
Added RD_PROFESSIONAL_API_URL

JackReevies:


Itay-Kirsh (approved):

apps/et/et-cos/demo.yaml
```diff
@@ -20,6 +20,8 @@ spec:
         NOC_NOTIFICATION_TEMPLATE_OLD_REP_ID: fe52b39f-852c-43ca-a42a-b9a27c43b130
         NOC_NOTIFICATION_TEMPLATE_NEW_REP_ID: 8fe52f24-40b2-4986-8f86-4dd1af311cbd
         NOC_NOTIFICATION_TEMPLATE_TRIBUNAL_ID: 1d5efcbd-1971-4ebe-bfe8-72ba36b5abac
+        RD_PROFESSIONAL_API_URL: http://rd-professional-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
```
GauravChawlaMOJ, line 23: As we are adding this only for demo,  we can just write demo rather than {{ .Values.global.environment }} but it should work too :)

PR #22496, opened 2023-04-28
Update config  on Demo for stitching

yogesh-hullatti:
https://tools.hmcts.net/jira/browse/EM-5225
### Change description ###
Update config  on Demo for stitching
```
[ ] Yes
[X] No
```

### DIFF
--- a/apps/em/em-stitching/demo.yaml
+++ b/apps/em/em-stitching/demo.yaml
@@ -5,11 +5,6 @@ metadata:
 spec:
   values:
     java:
-      replicas: 1
-      memoryRequests: "3072Mi"
-      cpuRequests: "1500m"
-      memoryLimits: "6144Mi"
-      cpuLimits: "3000m"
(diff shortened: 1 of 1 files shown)


apps/em/em-stitching/demo.yaml
```diff
@@ -5,8 +5,14 @@ metadata:
 spec:
   values:
     java:
+      replicas: 1
+      memoryRequests: "3072Mi"
+      cpuRequests: "1500m"
+      memoryLimits: "6144Mi"
```
mdayican, line 11: I think we do not need these

mdayican (dismissed):

apps/em/em-stitching/demo.yaml
```diff
@@ -5,8 +5,14 @@ metadata:
...
+      cpuLimits: "3000m"
```
mdayican, line 12: I think we do not need these

mdayican (approved):

PR #22646, opened 2023-05-10
DTSPO-13682 - Add postgres flex values ithc

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-13692

### Change description ###
Add values for postgres flexible server in ithc to check new db is working

```
[ ] Yes
[x] No
```

### DIFF
--- a/apps/camunda/camunda-api/ithc.yaml
+++ b/apps/camunda/camunda-api/ithc.yaml
@@ -7,3 +7,10 @@ spec:
     java:
       environment:
         WA_AUTO_CONFIGURE_TASKS_ENABLED: false
+        CAMUNDA_DB_HOST: "hmcts-camunda-v14-flexible-{{ .Values.global.environment }}.postgres.database.azure.com"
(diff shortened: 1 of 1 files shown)


apps/camunda/camunda-api/ithc.yaml
```diff
@@ -7,3 +7,10 @@ spec:
     java:
       environment:
         WA_AUTO_CONFIGURE_TASKS_ENABLED: false
+        CAMUNDA_DB_HOST: "hmcts-camunda-v14-flexible-{{ .Values.global.environment }}.postgres.database.azure.com"
+        CAMUNDA_DB_USER_NAME: "camundaadmin"
```
thomast1906, line 11: Are we planning to change the username from default postgresql user?

endakelly: The existing single servers for camunda use a similar username to this but with `@hmcts-camunda-$env`. The flexible module sets the username without the `@` but the inputs to the module are the same as before.

thomast1906 (approved):

PR #22647, opened 2023-05-10
DTSPO-13682 - Update postgres flex values ithc

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-13692

### Change description ###
Pods on ithc are running but show an error concerning app insights connection string
Adding other secrets from base chart to see if that fixes the error 

```
[ ] Yes
[x] No
```

### DIFF
--- a/apps/camunda/camunda-api/ithc.yaml
+++ b/apps/camunda/camunda-api/ithc.yaml
@@ -14,3 +14,9 @@ spec:
           secrets:
             - name: bpm-v14-POSTGRES-PASS
               alias: spring.datasource.password
+            - name: camunda-admin-password
+              alias: CAMUNDA_ADMIN_PASSWORD
+            - name: s2s-secret-camunda-bpm
+              alias: S2S_SECRET_CAMUNDA_BPM
+            - name: app-insights-connection-string
+              alias: app-insights-connection-string


apps/camunda/camunda-api/ithc.yaml
```diff
@@ -14,3 +14,9 @@ spec:
           secrets:
             - name: bpm-v14-POSTGRES-PASS
               alias: spring.datasource.password
+            - name: camunda-admin-password
+              alias: CAMUNDA_ADMIN_PASSWORD
+            - name: s2s-secret-camunda-bpm
+              alias: S2S_SECRET_CAMUNDA_BPM
+            - name: app-insights-connection-string
```
thomast1906, line 21: Is this not already in the base chart ?
https://github.com/hmcts/hmcts-charts/blob/master/stable/camunda-bpm/values.yaml#L14-L15

endakelly: It is yeah but the pods are suddenly complaining about the connection string: `Caused by: java.nio.file.NoSuchFileException: /mnt/secrets/camunda/app-insights-connection-string`. Just want to check if it's some issue with how the secrets are getting pulled in and mounted - can't see why it would be but the error wasn't present prior to the previous PR.

endakelly: Running `kubectl describe secretproviderclass camunda-api-java-camunda` against ithc shows that only `bpm-v14-POSTGRES-PASS` is being passed in whereas all the secrets are being passed in on demo. You would have thought it would combine the values from the base chart with the overriding values but seems not.

thomast1906 (approved):

thomast1906:
Approved to temporary test only. if it does resolve, we need to look at why base is not doing what it should

PR #22737, opened 2023-05-12
Flux-System: Add separate gotk-components.yaml for PTL, update kustomization for PTL

danielwilsonkainos:

https://tools.hmcts.net/jira/browse/DTSPO-12552

### Change description ###

- Add separate gotk-components.yaml for PTL including image automation components 
- Update kustomization.yaml to use new gotk-components.yaml and stop using base gotk-components.yaml without the image automation components
- Remove old image-automation-components.yaml file

```
[ ] Yes
[x] No
```


apps/flux-system/ptl-intsvc/base/kustomization.yaml
```diff
@@ -2,11 +2,16 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 namespace: flux-system
 resources:
-  - ../../base
-  - ../../base/image-automation-components.yaml
+  - ./gotk-components.yaml
```
adusumillipraveen, line 5: or move [this](https://github.com/hmcts/cnp-flux-config/blob/master/apps/flux-system/base/kustomization.yaml#L4) to all envs? ? Just conscious we are adding many files here seperately which could get missed when adding new ones in future ?

danielwilsonkainos: yeah i see what you're saying here. I think you're right and this may be a better implementation. I'll switch to this method.

timja (approved):

PR #22956, opened 2023-05-18
Removing checks on master

adusumillipraveen:
checks run on master add no values, given most of the commits are from image automation controller and not monitored.

```
[ ] Yes
[x] No
```


thomast1906 (approved):
true if PR covers.

PR #23023, opened 2023-05-23
Updated redaction whitelist on Prod

yogesh-hullatti:
https://tools.hmcts.net/jira/browse/EM-5280
### Change description ###
Updated redaction whitelist on Prod.
```
[ ] Yes
[X] No
```

### DIFF
--- a/apps/em/em-npa/prod.yaml
+++ b/apps/em/em-npa/prod.yaml
@@ -8,3 +8,4 @@ spec:
       environment:
         IDAM_API_BASE_URI: "https://idam-api.platform.hmcts.net"
         OPEN_ID_API_BASE_URI: "https://hmcts-access.service.gov.uk/o"
+        S2S_NAMES_WHITELIST: sscs,divorce,ccd,em_gw,xui_webapp


Jefex7 (approved):

apps/em/em-npa/prod.yaml
```diff
@@ -8,3 +8,4 @@ spec:
       environment:
         IDAM_API_BASE_URI: "https://idam-api.platform.hmcts.net"
         OPEN_ID_API_BASE_URI: "https://hmcts-access.service.gov.uk/o"
+        S2S_NAMES_WHITELIST: sscs,divorce,ccd,em_gw,xui_webapp
```
mdayican, line 11: why do not we change this for env?

yogesh-hullatti: Details in this ticket https://tools.hmcts.net/jira/browse/EM-5280
It is good practice to first ensure current config is replicated on Prod. Then move to change the config on non-prod environments. Once testing is completed, we will revert the config override from Prod file.

mdayican: you can change the config in test env first?

PR #23072, opened 2023-05-25
DTSPO-13897 - Override batch history ttl in prod

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-13897

### Change description ###
Add override value for batch history ttl
New value being set in updated chart
See https://github.com/hmcts/camunda-bpm/pull/550

```
[ ] Yes
[x] No
```

### DIFF
--- a/apps/camunda/camunda-ui/prod.yaml
+++ b/apps/camunda/camunda-ui/prod.yaml
@@ -6,3 +6,5 @@ spec:
   values:
     java:
       ingressHost: camunda-bpm.platform.hmcts.net
+      environment:
+        CAMUNDA_BATCH_HISTORY_TTL: "P30D"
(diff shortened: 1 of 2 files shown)


timja (commented):
do we want to override the schedule as well to be in the early hours?

apps/camunda/camunda-ui/prod.yaml
```diff
@@ -6,3 +6,5 @@ spec:
   values:
     java:
       ingressHost: camunda-bpm.platform.hmcts.net
+      environment:
+        CAMUNDA_BATCH_HISTORY_TTL: "P30D"
```
timja, line 10: probably want different schedules between these 2 instances of camunda

endakelly: What is the actual difference between these two apps? My understanding is that UI is just the frontend but they use the same image. Just not sure if we should make them different. When I updated the DB_HOST for BPM in other environments and didn't change it for UI so there was a discrepancy between the two which ended up causing issues.

timja (approved):

PR #23147, opened 2023-05-31
Revert "Temporarily  disable AAT Camunda pods"

kunlebalogun:
Reverts hmcts/cnp-flux-config#23132

Vacuum job completed


kunlebalogun:
Enabled Camunda on AAT following long vacuum purge

PR #23194, opened 2023-06-05
add civil-service into the whitelist

LeonardoPalmeiro:
https://tools.hmcts.net/jira/browse/CIV-8643

### Change description ###
Add the civil-service into the whitelist in order to fix  "status" : 403,   "error" : "Access Denied" from civil-service

```
[ ] Yes
[x] No
```


sampankumar (approved):

mdayican (changes requested):
civil team should access through CDAM

PR #23200, opened 2023-06-05
Add service bus secret for et

dharmendrak:


apps/et/preview/sops-secrets/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - et-sb-preview.enc.yaml
+  - et-sb-preview.enc.yaml
```
robin-ieong, line 5: seems to be duplicated?

PR #23377, opened 2023-06-15
Update FIC to create one per cluster

reespozzi:
This is to try and create separate resources per cluster so that we can account for the different oidc issuer urls per cluster

```
[ ] Yes
[ ] No
```


reespozzi:
Not too sure how this will work when both using the same managed identity, it may be smart enough to _just work_ - but want to test this

PR #23379, opened 2023-06-15
Fix for EXUI-343 to be deployed on Demo-int-1

kasi-subramaniam:
demo-int-1  needs this  PR  https://github.com/hmcts/rpx-xui-webapp/pull/3079 to be deployed


ernestman28 (approved):
lgtm

PR #23787, opened 2023-07-10
scale up pods and remove average utilizations

codaimaster:
-

### Change description ###
Scale up pods (increase memory request and limit) and remove average utilizations (use defaults)

```
[ ] Yes
[X] No
```

### DIFF
--- a/apps/xui/xui-webapp/prod.yaml
+++ b/apps/xui/xui-webapp/prod.yaml
@@ -8,13 +8,6 @@ spec:
       replicas: 24
       memoryLimits: "3Gi"
       memoryRequests: "1.5Gi"
-      autoscaling:
-        enabled: false
-        maxReplicas: 24
-        cpu:
-          averageUtilization: 70
-        memory:
-          averageUtilization: 70
       ingressHost: manage-case.platform.hmcts.net
       environment:
         HEALTH_CCD_COMPONENT_API: https://gateway.ccd.platform.hmcts.net/health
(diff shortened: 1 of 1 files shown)


andywilkinshmcts (changes requested):
1) I don't think it's a good idea to make this change on prod without testing. 
2) 5gb is a big heap, and a big increase. I think it would be better to  try with 4gb first
3) I think it's a bad idea to specify cpuLimits - it prevents the excess between limits and requests from being available to other pods in the container. You should just have cpuRequests - but I don't know what a suitable value for this would be - you'd have to look at what each pod is actually using

codaimaster:
> 1. I don't think it's a good idea to make this change on prod without testing.
> 2. 5gb is a big heap, and a big increase. I think it would be better to  try with 4gb first
> 3. I think it's a bad idea to specify cpuLimits - it prevents the excess between limits and requests from being available to other pods in the container. You should just have cpuRequests - but I don't know what a suitable value for this would be - you'd have to look at what each pod is actually using

CPU scale up looks indeed unnecessary but mem usage is saturating gradually. So, I'll suggest keeping CPU as is but propose a higher memory. 4gb is fine.

andywilkinshmcts (approved):

cakeben:
I will approve but I don't think it's the correct path to go down, we are not seeing OOMKills on the pods

cakeben (approved):

PR #23812, opened 2023-07-11
AM-2853: Enable dummy env var in orm prod config to allow prod pods to refresh

tomxelliott:


mattnayler:
Closing as emergency pods refresh was not required.

PR #24297, opened 2023-08-14
Switch to azure ad plugin

viktoriyabogdanova:
DTSPO-10435

### Change description ###
Switch CFT Jenkins to use azure ad plugin

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/apps/jenkins/jenkins/jenkins.yaml
+++ b/apps/jenkins/jenkins/jenkins.yaml
@@ -106,6 +106,14 @@ spec:
             unclassified:
               gitHubConfiguration:
                 apiRateLimitChecker: ThrottleOnOver
+          authentication: |
+            jenkins:
+              securityRealm:
+                azure:
+                  cacheDuration: 3600
+                  clientId: "f51ea833-3992-44a6-899e-16257b3e4fd1"
+                  clientSecret: "${cft-jenkins-sp-secret}"
+                  tenant: "531ff96d-0ae9-462a-8d2d-bec7c0b42082"
           auth: |
             jenkins:
               authorizationStrategy:
--- a/apps/jenkins/jenkins/jenkins.yaml
+++ b/apps/jenkins/jenkins/jenkins.yaml
@@ -117,25 +117,23 @@ spec:
           auth: |
             jenkins:
               authorizationStrategy:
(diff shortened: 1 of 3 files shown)


timja (approved):

apps/jenkins/jenkins/jenkins.yaml
```diff
@@ -106,6 +106,14 @@ spec:
...
                 apiRateLimitChecker: ThrottleOnOver
+          authentication: |
+            jenkins:
+              securityRealm:
+                azure:
+                  cacheDuration: 3600
+                  clientId: "f51ea833-3992-44a6-899e-16257b3e4fd1"
+                  clientSecret: "${cft-jenkins-sp-secret}"
+                  tenant: "531ff96d-0ae9-462a-8d2d-bec7c0b42082"
           auth: |
```
timja, line 117: you also need to update this as well

timja (changes requested):

timja (changes requested):
also need to remove the config from https://github.com/hmcts/cnp-flux-config/blob/a3f7fedeb7fa99076a00ef955a77916e09b26444/apps/jenkins/jenkins/ptl-intsvc/jenkins.yaml#L365 otherwise it will be overridden

apps/jenkins/jenkins/ptl-intsvc/jenkins.yaml
```diff
@@ -365,17 +365,31 @@ spec:
           authentication: |
```
timja, line 365: I think you can just remove this from PTL and move it to the base file?

apps/jenkins/jenkins/sbox-intsvc/jenkins.yaml
```diff
@@ -122,7 +122,7 @@ spec:
                 azure:
                   cacheDuration: 3600
                   clientId: "7e2f8c93-7d53-4668-bd84-f134d67dee09"
-                  clientSecret: "$${jenkins-client-secret}"
+                  clientSecret: "${jenkins-client-secret}"
```
timja, line 125: this should be removed from sbox as you've added the same config to the base now

timja (approved):

PR #24330, opened 2023-08-15
Move aso secrets to plain secrets

adusumillipraveen:
- They are not secrets and easier to handle in bootstrap if not managed in sops

```
[ ] Yes
[x] No
```


thomast1906 (dismissed):
Nice, this was something I was curious on at the time when we had a preview ASO issue.

PR #24540, opened 2023-08-29
DTSPO-14034 - Re-add arm daily pool

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-14034

SecurityScans are now fixed


cpareek (approved):
LGTM

PR #24590, opened 2023-08-31
Reduce number of ExUI replicas in prod.yaml to 20

andywilkinshmcts:
There is some evidence that ExUI is saturating redis. Reducing the number of instances to 20 may improve this

https://tools.hmcts.net/jira/browse/EXUI-775

### Change description ###

Reduce replica count from 24 to 20

```
[ ] Yes
[ x] No
```


codaimaster (approved):
Looks ok. But let's wait for the platform and performance metrics to stabilize after the fix of current P2.

PR #24741, opened 2023-09-08
Update script to set WI_NAME for new namespace

reespozzi:
https://tools.hmcts.net/jira/browse/DTSPO-13638

Updates script to add WI_NAME to their init namespace file. If product flag is passed, this is used, but it will default to namespace value.
Script switched to using flags.

```
[ ] Yes
[ ] No
```

### DIFF
--- a/bin/v2/add-namespace.sh
+++ b/bin/v2/add-namespace.sh
@@ -1,13 +1,40 @@
 #!/usr/bin/env bash
-set -ex
-
-NAMESPACE=$1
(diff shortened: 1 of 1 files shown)


reespozzi:
kind Kustomization
Example where product was set

adusumillipraveen (dismissed):

adusumillipraveen (dismissed):

adusumillipraveen (approved):

PR #24787, opened 2023-09-11
add filter for negate neuvector namespace

athomson-dt:

https://tools.hmcts.net/jira/browse/APM-2219

### Change description ###

Filtering out neuvector pods so they are not monitored by dynatrace this is a test for production but this monitoring is also not needed in AAT

```
[ ] Yes
[x] No
```

### DIFF
--- a/apps/dynatrace/aat/base/dynakube.yaml
+++ b/apps/dynatrace/aat/base/dynakube.yaml
@@ -9,3 +9,9 @@ spec:
     cloudNativeFullStack:
       args:
         - --set-host-group=AAT_CFT
+  namespaceSelector:
+     matchExpressions:
+     - key: helm.toolkit.fluxcd.io/namespace
+       operator: NotIn
+       values:
+       - neuvector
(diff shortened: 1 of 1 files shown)


apps/dynatrace/aat/base/dynakube.yaml
```diff
@@ -4,8 +4,14 @@ metadata:
...
-  apiUrl: https://yrk32651.live.dynatrace.com/api
+  apiUrl: 'https://yrk32651.live.dynatrace.com/api'
   oneAgent:
     cloudNativeFullStack:
       args:
-        - --set-host-group=AAT_CFT
+        - '--set-host-group=AAT_CFT'
+  namespaceSelector:
+    matchExpressions:
+      - key: helm.toolkit.fluxcd.io/namespace
```
timja, line 14: I would use the actual namespace name, copied from https://github.com/hmcts/sds-flux-config/blob/ad1da12cbc9de4a797a476220f4f71505f832d36/apps/dynatrace/dynakube.yaml#L68-L73
```suggestion
      - key: kubernetes.io/metadata.name
```

athomson-dt: Yeah agree with that, updated

timja (approved):

thomast1906 (approved):

PR #24828, opened 2023-09-12
enable wi on preview

viktoriyabogdanova:
DTSPO-13637

### Change description ###
Enable plum workload identity for Preview

```
[ ] Yes
[ x ] No
```


apps/cnp/plum-frontend/preview.yaml
```diff
@@ -0,0 +1,9 @@
+---
```
timja, line 1: plum doesn't run on preview as a permanent app only for pull requests, it doesn't make sense to add this at all

PR #24864, opened 2023-09-13
Switch to post build rather than patches for TT

timja:
So that we don't need to duplicate the ID in every environment between SA and app,

see:
https://github.com/hmcts/cnp-flux-config/blob/f3dea62290c5a4ea5dd38a2a6f16e237ac440412/apps/cnp/plum-recipe-backend/plum-recipe-backend.yaml#L13
and
https://github.com/hmcts/cnp-flux-config/blob/f3dea62290c5a4ea5dd38a2a6f16e237ac440412/apps/cnp/aat/base/kustomize.yaml#L9

No flux config for TT app yet so not fully complete


apps/ccd/ccd-int/ccd-int.yaml
```diff
@@ -136,7 +136,7 @@ spec:
         setup:
           topics:
             - name: ccd-case-events
-              subscriptionNeeded: yes
+              subscriptionNeeded: 'yes'
```
timja, line 139: this was causing a lint warning

PR #25125, opened 2023-09-25
added tags

cpareek:
### Change description ###
Added missing tags

```
[ ] Yes
[X ] No
```

### DIFF
--- a/apps/azureserviceoperator-system/resources/flexibleserver-postgres.yaml
+++ b/apps/azureserviceoperator-system/resources/flexibleserver-postgres.yaml
@@ -25,4 +25,3 @@ spec:
     key: PASSWORD
   storage:
     storageSizeGB: 32
-    
--- a/apps/azureserviceoperator-system/resources/flexibleserver-postgres.yaml
+++ b/apps/azureserviceoperator-system/resources/flexibleserver-postgres.yaml
@@ -6,6 +6,11 @@ metadata:
   annotations:
     serviceoperator.azure.com/reconcile-policy: detach-on-delete
 spec:
+  tags:
+    builtFrom: https://github.com/hmcts/cnp-flux-config
+    businessArea: CFT
+    environment: development
+    autoShutdown: true
   location: uksouth
   owner:
     name: ${NAMESPACE}-aso-${ENVIRONMENT}-rg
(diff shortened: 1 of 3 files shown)


apps/azureserviceoperator-system/resources/flexibleserver-postgres.yaml
```diff
@@ -6,6 +6,12 @@ metadata:
   annotations:
     serviceoperator.azure.com/reconcile-policy: detach-on-delete
 spec:
+  tags:
+    builtFrom: https://github.com/hmcts/cnp-flux-config
+    businessArea: CFT
+    environment: development
+    autoShutdown: "true"
+    application: social-service-child-support
```
timja, line 14: hmm this looks like a generic resource, it probably needs an environment variable being added into the environments post build setup.

e.g. in here:
https://github.com/hmcts/cnp-flux-config/blob/master/apps/sscs/base/kustomize.yaml
adding an APPLICATION_TAG in there

apps/sscs/preview/aso/sscs-postgres.yaml
```diff
@@ -6,6 +6,12 @@ metadata:
  annotations:
   serviceoperator.azure.com/reconcile-policy: detach-on-delete
 spec:
+ tags:
```
timja, line 9: why do we have the generic version in apps/azureserviceoperator-system/resources/flexibleserver-postgres.yaml and then this one here?

apps/azureserviceoperator-system/resources/flexibleserver-postgres.yaml
cpareek: I have added this

apps/sscs/preview/aso/sscs-postgres.yaml
cpareek: I think the ASO resource will need to setup in the apps/azureserviceoperator-system/resources first and then in the environement/app you want it.  Again, followed whats been done on service bus

timja: i don't think that's right, sscs should just add the resource to their preview kustomize file and then they will get this.

timja: although people probably want a bit of configuration available? i.e. database version

PR #25198, opened 2023-09-28
added new tag

cpareek:
Exempt from auto locking tags added to preview ASO resource group

### DIFF
--- a/apps/azureserviceoperator-system/resources/resource-group.yaml
+++ b/apps/azureserviceoperator-system/resources/resource-group.yaml
@@ -13,3 +13,4 @@ spec:
     businessArea: CFT
     environment: development
     exemptFromAutoLock: "true"
+    application: "${APPLICATION_TAG:-not-set-yet}"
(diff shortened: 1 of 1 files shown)


apps/azureserviceoperator-system/resources/resource-group.yaml
```diff
@@ -7,4 +7,9 @@ metadata:
...
 spec:
   location: uksouth
-  azureName: ${NAMESPACE}-aso-${ENVIRONMENT}-rg
\ No newline at end of file
+  azureName: ${NAMESPACE}-aso-${ENVIRONMENT}-rg
+  tags:
+    builtFrom: https://github.com/hmcts/cnp-flux-config
+    businessArea: CFT
+    environment: development
+    exemptFromAutoLock: "true"
```
timja, line 15: should probably add the application tag too with the same fallback as in the other one?

cpareek: sure -done

timja (approved):

PR #25237, opened 2023-09-29
Add azure extensions for postgres

dharmendrak:


apps/chart-tests/azure-service-operator/aso.yaml
```diff
@@ -60,4 +60,18 @@ spec:
     name: chart-test-postgres-password
     key: PASSWORD
   storage:
-    storageSizeGB: 32
\ No newline at end of file
+    storageSizeGB: 32
+---
+apiVersion: dbforpostgresql.azure.com/v1api20210601
```
timja, line 65: do you need to change the chart-tests namespace?

If so I would suggest a separate patch for this rather than multiple files in the same document I think that doesn't work properly any more

timja (approved):

timja (approved):

fergusoshea (approved):

PR #25286, opened 2023-10-02
Add wl identity docs

adusumillipraveen:
- will refer in chart-java release notes when merged
-  Replaced earlier docs as no new namespaces should use that.

### DIFF
--- a/docs/app-deployment-v2.md
+++ b/docs/app-deployment-v2.md
@@ -25,11 +25,11 @@ All the applications owned by a team are deployed to a single namespace (usually
     ./bin/v2/add-namespace-to-env.sh <your namespace> <environment>
    ```
 
-## WorkLoad Identity
+## Workload Identity
 
 - We use Managed Identity to access keyvaults secrets in application. This should be created after namespace is created above.
 - Namespace creation steps above creates a service account with the same name as namespace.
-- We need to add the managed identity client id as an annotation to service account to use work load identity.
+- We need to add the managed identity client id as an annotation to service account to use Workload Identity.
 
 ### Add WorkLoad identity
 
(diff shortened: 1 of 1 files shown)


docs/app-deployment-v2.md
```diff
@@ -25,30 +25,25 @@ All the applications owned by a team are deployed to a single namespace (usually
     ./bin/v2/add-namespace-to-env.sh <your namespace> <environment>
    ```
 
-## Managed Identity
+## WorkLoad Identity
 
-We use Managed Identity to access keyvaults secrets in application. This should be created after namespace is created above.
+- We use Managed Identity to access keyvaults secrets in application. This should be created after namespace is created above.
+- Namespace creation steps above creates a service account with the same name as namespace.
+- We need to add the managed identity client id as an annotation to service account to use work load identity.
```
timja, line 32:
```suggestion
- We need to add the managed identity client id as an annotation to service account to use Workload Identity.
```

docs/app-deployment-v2.md
```diff
@@ -25,30 +25,25 @@ All the applications owned by a team are deployed to a single namespace (usually
...
  ```bash
-    ./bin/v2/add-identity-to-env.sh <your namespace> <team short name> <mi name> <environment>
-    #example
-    ./bin/v2/add-identity-to-env.sh divorce div div aat
+    ./bin/workload-identity/add-wl-identity.sh --namespace <your namespace> --mi-name <mi name>
+    #examples
+    ./bin/workload-identity/add-wl-identity.sh --namespace ccd
+    ./bin/workload-identity/add-wl-identity.sh --namespace rpe --mi-name rpe-shared
    ```
+- In cases where u need to access other application key vaults (like in preview), your [Managed Identity should added to respective vault](https://github.com/hmcts/ccd-shared-infrastructure/pull/178/files) in specific environment.
```
timja, line 46:
```suggestion
- In cases where you need to access other application key vaults (like in preview), your [Managed Identity should be added to the respective vault](https://github.com/hmcts/ccd-shared-infrastructure/pull/178/files) in specific environment.
```

timja (changes requested):
typos

timja (approved):

PR #25377, opened 2023-10-04
DTSPO-15402  increasing max connection for postgres server

cpareek:
Increasing max connection for the postgres server to solve this error - 
https://tools.hmcts.net/jira/browse/DTSPO-15402


cpareek:
> I don't think this is the issue, we already bumped this to 3000 and it didn't help and anyway it shouldn't be getting anywhere near this

Yes  - I think when last time is bumped to 3K, the test wasn't done correctly.  This time, I had first changed it on portal, waited for it to finish the restart and then connected to the DB and then asked Dharemendra to run his pipeline.  He has confirmed his pods are up and running now.

I think the problem might be that the apps are opening the connection but not closing it correctly which then leaving the connection open for short time.

timja (approved):

PR #25483, opened 2023-10-10
pr-3277 PR to deploy on demo-int.

kasi-subramaniam:
pr-3277 PR to deploy on demo-int.


apps/xui/xui-webapp-integration/demo-image-policy.yaml
kasi-subramaniam, line 1: in demo.yaml the image: entry continues to point to prod. the demo-image-policy-file.yaml has the reference to the pr-3277 and this should suffice for correct deployment

danlysiak (approved):

PR #25660, opened 2023-10-17
Update kustomization.yaml

reespozzi:
Seems other envs have ../../base instead besides preview

sbox not working correctly

### DIFF
--- a/apps/rpe/sbox/base/kustomization.yaml
+++ b/apps/rpe/sbox/base/kustomization.yaml
@@ -1,7 +1,8 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - ../../base
+  - ../../../base
+  - ../../../base/workload-identity
   - ../../draft-store-service/draft-store-service.yaml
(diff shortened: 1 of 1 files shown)


apps/rpe/sbox/base/kustomization.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - ../../../base
+  - ../../base
```
adusumillipraveen, line 4: Might be intentional ?

reespozzi: Do you know potential reason? RPE namespace isn't created managed identity - federated credentials etc

adusumillipraveen: Just add workload identity like preview i think. Usually we dont add everything in base to preview and sbox to avoid deploying unnecessary apps.

reespozzi: Ah of course - done

PR #25791, opened 2023-10-23
Enable Global search tab for employment in prod

dharmendrak:


andywilkinshmcts (approved):
I've discussed this with Dharmendra and checked it is working on AAT.

PR #25983, opened 2023-10-30
SSCS kustomize config was commented previously as some secrets were missing

nitinprabhuhmcts:


thomast1906 (approved):
Confirmed with Nitin that secret is now there in ITHC

PR #26033, opened 2023-11-01
DTSPO-15344: Add create-job role to rd nonprod

reespozzi:
Adds role that allows rd to create jobs and cronjobs in nonprod

### DIFF
--- a/apps/rbac/create-job-role.yaml
+++ b/apps/rbac/create-job-role.yaml
@@ -8,10 +8,7 @@ rules:
 - apiGroups: ["batch"]
   resources: ["jobs"]
   verbs: ["create"]
-- apiGroups: ["batch"]
-  resources: ["cronjobs"]
-  verbs: ["create"]
-
+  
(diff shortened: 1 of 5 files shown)


apps/rbac/create-job-role.yaml
```diff
@@ -0,0 +1,28 @@
...
+metadata:
+  name: create-job-role
+  namespace: ${NAMESPACE}
+rules:
+- apiGroups: ["batch"]
+  resources: ["jobs"]
+  verbs: ["create"]
+- apiGroups: ["batch"]
+  resources: ["cronjobs"]
+  verbs: ["create"]
```
timja, line 13: why? the request was just for jobs

reespozzi: There was also many mentions of cronjobs in that ticket as well, so I added in case both were needed
Will remove though

timja (approved):

PR #26987, opened 2023-11-28
DTSPO-15525: Update svc type to loadbalancer for multi-cluster

reespozzi:
https://tools.hmcts.net/jira/browse/DTSPO-15525

Have tried this via clusterIP behind traefik and over port 80, but it explicitly needs these services exposing over these ports
```
There MUST be network connectivity between the controllers in each cluster on the required ports. The controller is exposed external to its cluster by either a primary or remote service, as can be seen in the sample NeuVector deployment yaml file. 
```


reespozzi:
> you probably want to assign a specific IP so that it works on rebuilds?

Just seeing if I can add that through the chart somehow

reespozzi:
> you probably want to assign a specific IP so that it works on rebuilds?

It's not currently possible in the chart
Have made a PR upstream, but will edit our local chart in the same way so we can define an IP

timja (approved):

PR #27119, opened 2023-12-01
update sptribs cron configurations

HarryH96:
### Jira link (if applicable)
N/A

### Change description ###
- Bumping sptribs-cron to reference updated chart version
- Adding suspend params for migrate cron jobs
- Configuring PROD crons to run out of hours (also currently suspended so they should not trigger in prod)


sjamihmcts (dismissed):

colmlynch60 (approved):
LGTM

PR #27463, opened 2023-12-14
HwF - Ithc setup

zaparka:


apps/help-with-fees/serviceaccount/ithc.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: ${NAMESPACE}
+  namespace: ${NAMESPACE}
+  annotations:
+    azure.workload.identity/client-id: ""
```
thomast1906, line 7: you probably need to have a value here if you are wanting to use that?

zaparka: @thomast1906 I'm in the process of finding the value for it.

thomast1906: Sure - we will hold off merging until value added

garytaylor (approved):

apps/help-with-fees/serviceaccount/ithc.yaml
zaparka: @thomast1906 ready to merge

timja (approved):

PR #27476, opened 2023-12-14
St add idamurl config

osheaf01:
### Jira link (if applicable)

DSS Create Case release

### Change description ###

Configuring IDAM URLs


colmlynch60 (approved):
LGTM

PR #28259, opened 2023-12-20
Upgrade env injector

timja:


thomast1906 (approved):

apps/admin/env-injector/env-injector.yaml
```diff
@@ -14,7 +14,7 @@ spec:
         name: hmctspublic
         namespace: flux-system
   values:
-    image: hmctspublic.azurecr.io/hmcts/k8s-env-injector:496359_20231218
+    image: hmctspublic.azurecr.io/hmcts/k8s-env-injector:497670_20231220
```
thomast1906, line 17: Just a thought - is it worth aligning the tag to our standard tagging policy?

timja: maybe but not right now

PR #28744, opened 2024-01-19
ST-1728 Point Demo at R2/2.1

osheaf01:
### Change description ###
Pointing sptribs-case-api on Demo to PR1384


colmlynch60 (approved):
LGTM

PR #28862, opened 2024-01-25
NFDIV-3855 - Example move to cron-chart 2.1.1 config

amoriartyMOJ:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/NFDIV-3855

### Change description ###
This PR serves as an example to NFD developers on how to move to the new "Active Cluster" config. In cron-chart v2.0.0+ Platops made a change which meant we no longer need to manually manage our configs per cluster (e.g. 00.yaml / 01.yaml).

1. For each environments config, rename the 00.yaml to <env>.yaml (e.g. 00.yaml -> aat.yaml) and then delete the 01.yaml config.

2. Add an override to each new yaml config files job section to use disableActiveClusterCheck: false. This will allow us to have active cluster management handled by platops instead of manually setting it ourselves.

3. confirm the schedule is still appropriate for the cron; is it running too frequently? Not enough?

4. If a yaml file sets a specific chart version, delete it as our base yaml config per cron should set it to the correct version.

Tip:
Split your ticket into 2 sections. First make the change in config to only the lower test environments. Deploy it and make sure everything works fine in AAT. Then create a final PR to move prod to the new config.


DelythJustice (approved):

danielwilsonkainos (approved):
LGTM

PR #28884, opened 2024-01-25
DTSPO-16508 - upgraded non prod traefik

cpareek:
### Jira link (if applicable)

https://tools.hmcts.net/jira/browse/DTSPO-16508

### Change description ###
Upgraded traefik to v26, followed same configuration done as on sbox

### Checklist

### DIFF
--- a/apps/admin/traefik2/ptl-intsvc/00.yaml
+++ b/apps/admin/traefik2/ptl-intsvc/00.yaml
@@ -4,15 +4,6 @@ metadata:
   name: traefik2
   namespace: admin
 spec:
-  chart:
-    spec:
-      chart: traefik
-      # update the crd version in traefik-crds when updating this
-      version: 26.0.0
-      sourceRef:
-        kind: HelmRepository
-        name: traefik
-        namespace: admin
   values:
     service:
       spec:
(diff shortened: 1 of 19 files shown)


thomast1906 (approved):

apps/admin/traefik2/ptl-intsvc/00.yaml
```diff
@@ -4,6 +4,15 @@ metadata:
   name: traefik2
   namespace: admin
 spec:
+  chart:
```
endakelly, line 7: I would do ptl in sds first just to be sure of Jenkins working. Should be fine but best to keep it consistent

cpareek: ah -sure, I was thinking this would be different then ptl  environment as it said ptl-intsvc :)

endakelly: yeah, sadly they're not the same across both repos but they are the same use case

cpareek: sure

endakelly (approved):

endakelly (approved):

PR #29036, opened 2024-02-02
 PAY-6846: Civil Case: State is not changed to "Awaiting Notification…

Thor-tech-of-metal:
PAY-6846: Civil Case: State is not changed to "Awaiting Notification" after payment is successful
             [PAY-6846] (https://tools.hmcts.net/jira/browse/PAY-6846).


Ollie-Smith (approved):
reviewed

PR #29390, opened 2024-02-20
deschduler config for sbox

NickAzureDevops:
### Jira link (https://tools.hmcts.net/jira/browse/DTSPO-16716)

deschduler config for sbox

### Checklist


apps/cnp/sbox/base/descheduler.yaml
reespozzi, line 1: Can these files be renamed? General approach in this repo is more like
descheduler-helmrepo.yaml

apps/cnp/sbox/base/descheduler2.yaml
reespozzi, line 1: Can this be set to descheduler.yaml once the helmrepo file is renamed? Calling a file something2 isn't ideal

apps/cnp/sbox/base/descheduler2.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: helm.toolkit.fluxcd.io/v2beta2
+kind: HelmRelease
+metadata:
+  name: descheduler2
```
reespozzi, line 4: Similar idea with naming here

apps/cnp/sbox/base/descheduler2.yaml
```diff
@@ -0,0 +1,74 @@
...
+  namespace: cnp
+spec:
+  timeout: 20m
+  releaseName: descheduler
+  chart:
+    spec:
+      chart: descheduler
+      version: 0.18.0
```
reespozzi, line 12: Is there a reason for 0.18? Latest is 0.29.0 https://github.com/kubernetes-sigs/descheduler/blob/master/charts/descheduler/Chart.yaml#L3 -- don't want to test config as an example that might have been broken/changed in more recent releases

apps/cnp/sbox/base/descheduler2.yaml
```diff
@@ -0,0 +1,74 @@
...
+        memory: 256Mi
+    securityContext:
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop:
+          - ALL
+      privileged: false
+      readOnlyRootFilesystem: true
+      runAsNonRoot: true
+      runAsUser: 1000
```
reespozzi, line 35: I cant see this stuff in helm chart values? Looks like default code from the job spec. Can remove? There's probably a bunch more stuff that was in the YAML output of the resource but not configurable via values this way

reespozzi (changes requested):

apps/cnp/sbox/base/descheduler2.yaml
```diff
@@ -0,0 +1,74 @@
...
+        port: 10258
+        scheme: HTTPS
+      initialDelaySeconds: 3
+      periodSeconds: 10
+    service:
+      enabled: false
+    serviceMonitor:
+      enabled: false
+      insecureSkipVerify: true
+      serverName: null
```
reespozzi, line 74: Can you check which of these are defaults and remove duplicates? A lot of config to look through for stuff pre-defined upstream the same way

apps/cnp/sbox/base/cronjob.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: helm.toolkit.fluxcd.io/v2beta2
+kind: HelmRelease
+metadata:
+  name: cronjob
```
reespozzi, line 4: We still want this to be called descheduler?

reespozzi (approved):
Looks OK to me now - want to make sure other stuff installed manually is removed before merging this please so there's no conflicts?

PR #29493, opened 2024-02-27
SSCSCI-576/22 - Turn off cookie flags in prod

gemmatalbot:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/SSCSCI-576
https://tools.hmcts.net/jira/browse/SSCSCI-22

### Change description ###
Turn off the httpOnly, sameSite and secure cookie flags in prod for MYA and SYA

### DIFF
--- a/apps/sscs/sscs-cor-frontend/prod.yaml
+++ b/apps/sscs/sscs-cor-frontend/prod.yaml
@@ -16,4 +16,7 @@ spec:
         MEDIA_FILES_ALLOWED_ENABLED: "true"
         REQUEST_TAB_ENABLED: "true"
         ALLOW_COOKIE_BANNER_ENABLED: "true"
+        HTTPONLY_COOKIE_FLAG_ENABLED: "true"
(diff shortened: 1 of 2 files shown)


apps/sscs/sscs-cor-frontend/prod.yaml
```diff
@@ -16,7 +16,4 @@ spec:
         MEDIA_FILES_ALLOWED_ENABLED: "true"
         REQUEST_TAB_ENABLED: "true"
         ALLOW_COOKIE_BANNER_ENABLED: "true"
-        HTTPONLY_COOKIE_FLAG_ENABLED: "true"
```
benouaer, line 19: I'd set it to false rather than remove it, because if we do that, it means we still need to investigate issues
we can remove once this is proven to work, and we remove the flag from the code

KatKovacs1 (approved):

hashimalisolirius (approved):

benouaer (approved):

PR #29500, opened 2024-02-27
[NFDIV-3873] HPA configuration for nfdiv frontend and case-api (AAT test)

adamg-hmcts:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/NFDIV-3873

### Change description ###
The HPA configuration has been moved from individual service repos to flux as per the recommendation in the [HPA guidance doc](https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1753681181). After monitoring autoscaling performance and discussing it with the performance testing team, we've also increased memory requests slightly, from 1024 Mb to 1536 Mb (nfdiv-case-api) and 256 Mb to 512 Mb (nfdiv frontend), to reduce scaling under normal conditions. This is a test for the new configuration (non-production environment) and if performance is okay, we would hope to open another PR to apply the changes in production.

### DIFF
--- a/apps/nfdiv/nfdiv-case-api/aat.yaml
+++ b/apps/nfdiv/nfdiv-case-api/aat.yaml
@@ -5,6 +5,14 @@ metadata:
 spec:
   values:
     java:
+      replicas: 2
+      autoscaling:
+        enabled: true
+        maxReplicas: 9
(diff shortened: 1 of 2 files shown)


amoriartyMOJ (dismissed):
LGTM

PR #29502, opened 2024-02-27
Deploy sscs ccd case migration perftest

markbrooks725:
### Change description ###
Deploy sscs-ccd-case-migration to perftest


AndreiTrushkov (approved):

narinder17 (approved):

timja (commented):
Can you remove the 00 and 01 patches, I don't see a reason for them as they are just duplicating content

markbrooks725:
@timja We don't want the job running on both clusters so we use the 00 and 01 patches to control which cluster it runs on. I have already merged but can we do that without the 00 and 01 patches?

timja:
> @timja We don't want the job running on both clusters so we use the 00 and 01 patches to control which cluster it runs on. I have already merged but can we do that without the 00 and 01 patches?

Platform Ops control the cluster centrally you don't need to do anything for that it will just run on 00 by default

PR #29660, opened 2024-03-06
Remove sptribs-dss-update-case-web from base kustomization instead put in specific env kustomization files.

tomxelliott:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/DTSPO-17008

### Change description ###
Remove sptribs-dss-update-case-web from base kustomization instead put in specific env kustomization files.


jordankainos (approved):
lgtm

PR #30364, opened 2024-04-11
DTSPO-16593 - turn off idam on sbox to carry on upgrade'

cpareek:
### Jira link (if applicable)

https://tools.hmcts.net/jira/browse/DTSPO-16593

### Change description ###

idam-api and idam-web-public containers is not in running state for some reason on sbox, did try to fix but didn't work so turning it off to perform the Custer upgrade.

### DIFF
--- a/apps/idam/sbox/base/kustomization.yaml
+++ b/apps/idam/sbox/base/kustomization.yaml
@@ -6,8 +6,8 @@ resources:
   - ../../idam-user-profile-bridge/idam-user-profile-bridge.yaml
 namespace: idam
 patches:
-  - path: ../../idam-api/sbox.yaml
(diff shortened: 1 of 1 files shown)


thomast1906 (changes requested):
Are they not running currently? 

IDAM do use sbox alot. If pods are running fine - do let Paul Verity / Kremena know

cpareek:
> Are they not running currently?
> 
> IDAM do use sbox alot. If pods are running fine - do let Paul Verity / Kremena know

So the pod is in running state but container is not, will speak to Paul now
image

thomast1906:
Sure , give Paul a FYI and then should be ok to upgrade

thomast1906 (approved):
Cheers

PR #30383, opened 2024-04-11
1777 on demo

kasi-subramaniam:
Test the Fix for 1777 / PAY-6933 on demo - 

Update the FEE_AND_PAY_API url  to point to the FeePay URL as below. This need to be done for both MO and AO on demo environment.

FEE_AND_PAY_API: http://payment-api-int-demo.service.core-compute-demo.internal . This entry will be reversed once testing is complete for https://tools.hmcts.net/jira/browse/PAY-6933.


andywilkinshmcts (changes requested):
Also need to do it for AO

kasi-subramaniam:
> Also need to do it for AO

Andy , both AO and MO are referred in this PR itself  

> Also need to do it for AO

kasi-subramaniam:
> Also need to do it for AO

MC , MO and AO - are all changed now .

andywilkinshmcts (approved):

PR #30634, opened 2024-04-22
[ADOP-2323] changed ithc PR image

hfilmore:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/ADOP-2323

### Change description ###
Changed ITHC image as part of ADOP-2323 testing

### DIFF
--- a/apps/adoption/adoption-web/ithc-image-policy.yaml
+++ b/apps/adoption/adoption-web/ithc-image-policy.yaml
@@ -6,7 +6,7 @@ metadata:
     hmcts.github.com/prod-automated: disabled
 spec:
   filterTags:
-    pattern: '^pr-1515-[a-f0-9]+-(?P<ts>[0-9]+)'
+    pattern: '^pr-1508-[a-f0-9]+-(?P<ts>[0-9]+)'
     extract: '$ts'
   policy:
     alphabetical:


apps/adoption/adoption-web/ithc-image-policy.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
     hmcts.github.com/prod-automated: disabled
 spec:
   filterTags:
-    pattern: '^pr-1515-[a-f0-9]+-(?P<ts>[0-9]+)'
```
GauravChawlaMOJ, line 9: PR number has to be changed here as well:
https://github.com/hmcts/cnp-flux-config/blob/2155e0e2e2116c10d172844bdd3be4c6cb63982d/apps/adoption/adoption-web/ithc.yaml#L11

GauravChawlaMOJ (changes requested):

apps/adoption/adoption-web/ithc-image-policy.yaml
hfilmore: Hi Gaurav, thanks for the review!  The problem with relying on changing the image is I'd have to update the ithc.yaml every time the PR is updated. The regex is really nice because if I push a new commit to the PR, the image gets updated automatically.  (The only time I changed the image was because I got into a muddle with merge conflicts!)  Hope that makes sense?  Message me if not.

PR #30766, opened 2024-04-29
use main branch for DEMO

mdayican:
### Jira link (if applicable)

use main branch for DEMO

## 🤖GIPPI PR SUMMARY🤖

- Updated `demo.yaml` in `em-hrs-api`:
  - Removed a line with environment configuration values for the Java image.

### DIFF
--- a/apps/em/em-hrs-api/demo.yaml
+++ b/apps/em/em-hrs-api/demo.yaml
@@ -5,6 +5,5 @@ metadata:
 spec:
   values:
     java:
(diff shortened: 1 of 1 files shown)


yogesh-hullatti (changes requested):
I suggest doing the change here - https://github.com/hmcts/cnp-flux-config/blob/master/apps/em/em-hrs-api/demo-image-policy.yaml
And leave demo.yaml as it is. Following the convention

PR #30939, opened 2024-05-07
field values updated

btensay:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/RET-4892

### Change description ###
thread state and cron job interval updated

### Checklist

- [x ] commit messages are meaningful and follow good commit message guidelines

### apps/et/et-ccd-case-migration/demo-00.yaml
- Updated the job schedule from \"*/5 * * * *\" to \"*/30 * * * *\"
- Changed the job suspension from true to false
- Increased the migration query size from 100 to 10000


apps/et/et-ccd-case-migration/et-ccd-case-migration.yaml
```diff
@@ -34,4 +34,4 @@ spec:
         kind: GitRepository
         name: hmcts-charts
         namespace: flux-system
-      interval: 1m
+      interval: 30m
```
hsjhita, line 37: This shouldn't change

btensay: Reverted

PR #32387, opened 2024-06-14
NFDIV-4084 Update crons config in PROD

pallavijustice:
Jira link (if applicable)

https://tools.hmcts.net/jira/browse/NFDIV-4084

Change description
NFDIV-4084 Update crons config in PROD

Checklist
 commit messages are meaningful and follow good commit message guidelines
 README and other documentation has been updated / added (if needed)


apps/nfdiv/nfdiv-cron-bulk-list-create/prod-00.yaml
```diff
@@ -1,27 +0,0 @@
-apiVersion: helm.toolkit.fluxcd.io/v2
-kind: HelmRelease
-metadata:
-  name: nfdiv-cron-bulk-list-create
-spec:
-  releaseName: nfdiv-cron-bulk-list-create
-  values:
-    job:
-      schedule: "0,30 5,6 * * *"
```
adamg-hmcts, line 9: I wonder why we had 2 config files for this cron with different timings - prod-00.yaml and prod/00.yaml? It looks like the job timing is 30 mins in the logs and I'm wondering whether this is the config we have been using in prod

apps/nfdiv/nfdiv-cron-migrate-cases/prod/00.yaml
```diff
@@ -6,6 +6,7 @@ spec:
   releaseName: nfdiv-cron-migrate-cases
   values:
     job:
+      disableActiveClusterCheck: false
```
adamg-hmcts, line 9: Do we need to create prod.yaml for this cron?

apps/nfdiv/nfdiv-cron-bulk-list-create/prod-00.yaml
pallavijustice: I have noticed that for a number of crons, the historical prod 00/01 yaml files have been retained even when they are not being referenced. That is why I was deleting those files. I will retain the historical files as of now. May be delete unused files as part of another task later on.

apps/nfdiv/nfdiv-cron-migrate-cases/prod/00.yaml
pallavijustice: Yes I think I missed out on renaming the file to prod.yaml

PR #34112, opened 2024-09-04
Add Artifactory helm chart

JoshuaGlassKainos:
### Jira link
Tech Improvement to release Artifactory via helm charts

### Change description

added Artifactory-helm directory with respective yaml code to configure artifactory via helm
added helm file to sbox kustomize file
updated disk size to 128Gb and to standard only

Disabled postgres usage
Disabled nginx usage
`allowNonPostgresql` set to true
Set version to latest `7.90.9`
Deployment in place of current sbox deployment

### Checklist

## 🤖AEP PR SUMMARY🤖

- added new file `artifactory-helmrepo.yaml` with HelmRepository configuration for Artifactory
- added new file `artifactory.yaml` with HelmRelease configuration for Artifactory
- updated `kustomization.yaml` in `artifactory/base` to remove a file reference and add `artifactory.yaml`
- updated `kustomization.yaml` in `ptl-intsvc/base` to add `artifactory.yaml` and remove `ingress.yaml`
- updated `kustomization.yaml` in `sbox-intsvc/base` to add `artifactory-helmrepo.yaml` and `artifactory.yaml` and remove `sbox-intsvc-ingress.yaml`

### DIFF
--- a/apps/artifactory/artifactory-helm/artifactory.yaml
+++ b/apps/artifactory/artifactory-helm/artifactory.yaml
@@ -1,0 +1,38 @@
+apiVersion: helm.toolkit.fluxcd.io/v2
+kind: HelmRelease
+metadata:
+  name: artifactory-oss
+  namespace: artifactory
+spec:
+  timeout: 20m
+  releaseName: artifactory-oss
+  chart:
+    spec:
+      chart: jfrog/artifactory-oss
+      version: 7.90.9
+      sourceRef:
+        kind: HelmRepository
+        name: artifactory
+        namespace: artifactory
+  interval: 1m
+  values:
+    artifactory:
+      #speicifes the artifactory-oss container image version
+      image:
+        tag: 7.90.9
+      persistence:
+        enabled: true
+        storageClassName: "managed"
+        size: 128Gi
+      replicaCount: 2
+    database:
+      allowNonPostgresql: true
+    nginx:
+      enabled: false
+    ingress:
+      enabled: true
+      hosts:
+        - artifactory.sandbox.platform.hmcts.net
+      className: traefik
+      annotations:
+        traefik.ingress.kubernetes.io/router.tls: "true"
--- a/apps/artifactory/artifactory-helm/artifactory.yaml
+++ b/apps/artifactory/artifactory-helm/artifactory.yaml
@@ -27,6 +27,8 @@ spec:
       replicaCount: 2
     database:
       allowNonPostgresql: true
+    postgresql:
+      enabled: false
     nginx:
       enabled: false
     ingress:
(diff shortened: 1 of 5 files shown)


danielwilsonkainos (dismissed):

apps/artifactory/sbox-intsvc/base/kustomization.yaml
```diff
@@ -3,3 +3,5 @@ kind: Kustomization
 resources:
   - ../../base
   - ../../artifactory/sbox-intsvc-ingress.yaml
```
reespozzi, line 5: is the intention now to keep existing deployment and run this alongside til we know it works?
You'll have to add some DNS records for this new ingress as well if going down this route - and then knock on effects of cleanup after / changing this elsewhere

Worth just replacing?

MartyFox, line 5: You will need to remove this ingress patch as you've got the ingress settings included in your chart

reespozzi: Ingress in this PR is a slightly different host name `  - artifactory-helm.sandbox.platform.hmcts.net` - think it could deploy

danielwilsonkainos: think it depend how the ingress created by the chart is named/its metadata right? Much easier to avoid and just update in place as mentioned in comment above

apps/artifactory/artifactory-helm/artifactory.yaml
```diff
@@ -0,0 +1,40 @@
...
+        kind: HelmRepository
+        name: artifactory
+        namespace: artifactory
+  interval: 1m
+  values:
+    artifactory:
+      #speicifes the artifactory-oss container image version
+      image:
+        tag: 7.90.9
+      persistence:
```
reespozzi, line 23: Don't we need to define https://github.com/jfrog/charts/blob/master/stable/artifactory/values.yaml#L507 ? Thought that this is how we get a pvc by using the helm chart

reespozzi: https://github.com/jfrog/charts/blob/master/stable/artifactory/templates/artifactory-statefulset.yaml#L1615

This^ which would reflect what we have in flux atm

JoshuaGlassKainos: we can also define pic through this var

https://github.com/jfrog/charts/blob/master/stable/artifactory/values.yaml#L720

reespozzi: Doing it that way should handle everything we need by [provisioning the PV too ](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#volume-claim-templates)

reespozzi: The template ordering is a bit confusing, but [that resource](https://github.com/jfrog/charts/blob/master/stable/artifactory/templates/artifactory-statefulset.yaml#L1595) will get deployed with the stateful set if it's defined in values

danielwilsonkainos (approved):

PR #34878, opened 2024-10-15
DTSRD-3544 - scheduling delta load in AAT every day at 8 AM

rajkatla-hmcts:
### Change description

Ccheduling delta load in AAT every day at 8 AM


mikebrownccd (approved):
LGTM

mattnayler (approved):

PR #38284, opened 2025-04-29
DTSPO-24229 docmosis scaling

ieuanb74:
### Jira link
https://tools.hmcts.net/jira/browse/DTSPO-24229

### Change description
- Ticket raised by perf test team during Docmosis upgrade that suggests we may be able to scale down the number of pods we have running in production (currently 3)
- After monitoring production it does seem that utilisation is generally low but I have noticed that there are still pods restarting with `OOMKilled` in Production. 
- Enabling autoscaling in prod with min replicas of 2 and max of 5 to try and prevent this while still keeping deployed resources at a minimum most of the time

### Checklist

## 🤖AEP PR SUMMARY🤖

- `docmosis.yaml`:
  - Increased replicas to 2 and enabled autoscaling with a maximum of 5 replicas.
  - Updated memory and CPU requests and limits.

- `perftest.yaml`:
  - Updated memory requests and limits.

- `prod.yaml`:
  - Updated memory requests.

### DIFF
--- a/apps/docmosis/docmosis/docmosis.yaml
+++ b/apps/docmosis/docmosis/docmosis.yaml
@@ -18,10 +18,14 @@ spec:
     envFromSecret: docmosis-secret
     image: hmctsprivate.azurecr.io/docmosis:prod-9d7e8a0-807256 #{"$imagepolicy": "flux-system:docmosis"}
     java:
-      memoryRequests: '1024Mi'
-      cpuRequests: '1000m'
-      memoryLimits: '3072Mi'
-      cpuLimits: '2000m'
+      replicas: 2
+      autoscaling:
+        enabled: true
+        maxReplicas: 5
+      memoryRequests: "1024Mi"
+      cpuRequests: "1000m"
+      memoryLimits: "3072Mi"
+      cpuLimits: "2000m"
   chart:
     spec:
       chart: base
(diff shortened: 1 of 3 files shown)


thomast1906 (approved):

thomast1906 (changes requested):
hmm, why not perftest autoscale too? same as prod

ieuanb74:
> hmm, why not perftest autoscale too? same as prod

Yep was close to doing this initially but didn't know how useful it would be. Will do that now!

dawidstrozak (approved):

thomast1906:
> > hmm, why not perftest autoscale too? same as prod
> 
> Yep was close to doing this initially but didn't know how useful it would be. Will do that now!

Changes like this, id want in perftest too - so they can let us know if for example autoscaling is not great for Docmosis etc

thomast1906:
@ieuanb74 default it to all envs I think

ieuanb74:
> > > hmm, why not perftest autoscale too? same as prod
> > 
> > 
> > Yep was close to doing this initially but didn't know how useful it would be. Will do that now!
> 
> Changes like this, id want in perftest too - so they can let us know if for example autoscaling is not great for Docmosis etc

Makes sense!

dawidstrozak (approved):

PR #38716, opened 2025-05-28
Reverting PR-39709 CCD-6345 after sanity check

bharatkumarpatel1:
Reverting PR-39709 CCD-6345 after sanity check


dinesh1patel (approved):
LGTM

PR #39008, opened 2025-06-17
Update bs-cron-trigger-daily-checks.yaml

amstevenson:
Setting env var to enable bsp-cron-trigger by default, but restrict on AAT for BP checks

## 🤖AEP PR SUMMARY🤖

- Updated file: bp-cron-trigger-daily-checks.yaml
  - Added environment variable TRIGGER_ENABLED with value false

- Updated file: prod.yaml
  - Added environment TRIGGER_ENABLED with value true

- Updated file: bs-cron-trigger-daily-checks.yaml
  - Added environment variable TRIGGER_ENABLED with value true


josh-uk (changes requested):
Why is bulk print the only one being disabled? and why only in AAT

PR #40430, opened 2025-09-15
update sops doc to extract sops and encrypt again

thomast1906:
### Change description
update sops doc to extract sops and encrypt again


thomast1906:
@ieuanb74 thoughts?

cpareek (approved):

PR #40455, opened 2025-09-16
Add Acro cron job for private law

dharmendrak:
Add Acro cron job for PRL


apps/private-law/prl-cron-acro-upload/prod.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: helm.toolkit.fluxcd.io/v2
+kind: HelmRelease
+metadata:
+  name: prl-cron-acro-upload
+  namespace: private-law
+spec:
+  releaseName: prl-cron-acro-upload
+  values:
+    job:
+      environment:
```
SomeshRA, line 10: I suppose env ACRO_CRONJOB_ENABLED: false would be required

SomeshRA (approved):
I suppose environment variable - ACRO_CRONJOB_ENABLED: false - would be required on all env apart from demo.

SomeshRA (changes requested):
I suppose environment variable - ACRO_CRONJOB_ENABLED: false - would be required on all env apart from demo.

apps/private-law/prl-cron-acro-upload/demo/01.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: helm.toolkit.fluxcd.io/v2
```
SomeshRA, line 1: As, cron job is executed on 1 cluster so i believe 01.yaml is not required

PR #40527, opened 2025-09-22
Dfr 4078 div cfs replicas never scale down to minpods

hamzahtahirhmcts:
### Jira link
[DFR-4078](https://tools.hmcts.net/jira/browse/DFR-4078)

### Change description
- replicas set to 1.
- HPA now enabled with minReplicas: 1, maxReplicas: 2
- Targets are default 80% for both CPU and Memory.

Metrics show div-cfs is at around 530540 and 2m CPU.
HPA reports 52% memory utilisation. 
replicas is set to 1 so it can go to one pod when quiet and 2 for the max.

### Security Vulnerability Assessment ###

**CVE Suppression:** Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?
  * [ ] Yes
  * [ ] No

### Checklist

### DIFF
--- a/apps/divorce/div-cfs/ithc.yaml
+++ b/apps/divorce/div-cfs/ithc.yaml
@@ -12,3 +12,10 @@ spec:
         VAR_FV2: ithc1
       prometheus:
         enabled: true
+      replicas: 1
+      autoscaling:
+        enabled: true
+        minReplicas: 1
+        maxReplicas: 2
+      cpuRequests: "250m"
+      memoryRequests: "1536Mi"
--- a/apps/divorce/div-cfs/div-cfs.yaml
+++ b/apps/divorce/div-cfs/div-cfs.yaml
@@ -8,15 +8,8 @@ spec:
   values:
     java:
       replicas: 2
-      autoscaling:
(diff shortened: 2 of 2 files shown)


apps/divorce/div-cfs/ithc.yaml
```diff
@@ -12,3 +12,10 @@ spec:
         VAR_FV2: ithc1
       prometheus:
         enabled: true
+      replicas: 1
+      autoscaling:
+        enabled: true
+        minReplicas: 1
+        maxReplicas: 2
```
hamzahtahirhmcts, line 19: I don't see how we will need maxReplicas set to 4 for ITHC. Unless anyone else says otherwise?

apps/divorce/div-cfs/ithc.yaml
```diff
@@ -12,3 +12,10 @@ spec:
         VAR_FV2: ithc1
       prometheus:
         enabled: true
+      replicas: 1
```
so99y, line 15: remove replicas: 1. Looks like it is not needed

PR #41506, opened 2025-10-16
fix: fixup sbox jenkins.yaml to set environment to be "sbox" which li…

coling01:
fix: fixup sbox jenkins.yaml to set environment to be "sbox" which lines up with the subscription and its prefixes which use sbox rather than "sandbox"


coling01:
Not sure its worth the pain to fixup sandbox ... if it works ok for other then no worries will leave it alone

PR #42120, opened 2025-11-27
DTSCCI-3392: Deploy prod image to perftest

ruban72:
### Jira link

<!-- 
DTSCCI-3392

### Change description
DTSCCI-3392: Deploy prod image to perftest

### DIFF
--- a/apps/civil/civil-service/perftest-image-policy.yaml
+++ b/apps/civil/civil-service/perftest-image-policy.yaml
@@ -6,7 +6,7 @@ metadata:
     hmcts.github.com/prod-automated: disabled
 spec:
   filterTags:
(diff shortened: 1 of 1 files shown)


apps/civil/civil-service/perftest.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   timeout: 10m
   values:
     java:
-      image: hmctspublic.azurecr.io/civil/service:pr-7197-517e756-20251127080432 #{"$imagepolicy": "flux-system:perftest-civil-service"}
```
gmikerlemos: if you want to deploy the prod image it is probably best to only change the image name to civil-service, you don't need to change the image itself to prod-c2d00a4-...

PR #44373, opened 2026-04-16
Update ithc.yaml

MarkHerbertCGI:
restart ithc environment following gov notify key rotation

### https://tools.hmcts.net/jira/browse/CRS-58


rupalilanke (approved):
Changes reviewed. Approving now

PR #44593, opened 2026-04-28
Revert "ADOP-2724 Add Adoption rate limiting to ITHC"

hfilmore:
Reverts hmcts/cnp-flux-config#43758


chak-shing-lo-justice (approved):
lgtm

PR #45333, opened 2026-06-04
DFR-5006_Cron_Update_Contact_details-3

dawudgovuk:
### Jira link
https://tools.hmcts.net/jira/browse/DFR-5006


hamzahtahirhmcts (approved):

ptrelease (changes requested):
Just want to check two things:

Your one off demo yaml hard-codes the PR, but also point at the image policy, which you've changed back to latest. So I would expect latest to be used.  What do you think?
I don't like the idea of scheduling anything in Prod for 13:30 BST before you have tested in demo.

ptrelease (approved):
Approving as the Prod config no longer applies, the time has passed.

After discussions, the policy should point at latest.

I've advised that they policy should be changed if the one-off in demo is to use the 2960 image later.

PR #45479, opened 2026-06-10
Add postgres secrets to clear inactive DSS cron

banderous:
This adds the postgres Key Vault secret mappings to the SPTRIBS clear inactive DSS draft cases cron job.

The cron job currently starts without POSTGRES_HOST, POSTGRES_PORT, POSTGRES_NAME, POSTGRES_USERNAME and POSTGRES_PASSWORD, so the case API falls back to its local defaults and tries to connect to localhost:6432 during Flyway startup. The added mappings match the other SPTRIBS cron jobs that run the same case API image.

Validated by parsing the updated HelmRelease YAML locally.


scottbdavey1993 (approved):
happy from my side waiting for @oleon23 too

oleon23 (approved):
Good from me - need to followup and check the other crons too

PR #46500, opened 2026-07-28
NPD Prometheus alerting

MartyFox:
### Jira link
https://tools.hmcts.net/jira/browse/DTSPO-33407

### Change description
- Add alerts in code for Node Problem Detector from metrics surfaced by Azure managed NPD.
- Add test setup within prometheus to allow testing before alerts are added to flux for deployment
  - Alert files have a test file that tests each alert as required
  - Script to run all tests within the testing setup
  - Yaml list for matching pairs (rule and test file)


apps/monitoring/kube-prometheus-stack/alert-testing/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Ignore the extracted native rules files - they are regenerated from the
+# PrometheusRule CRDs on every run of extract-and-test.sh.
```
DanielKChristou: The script is called `test.sh`

apps/monitoring/kube-prometheus-stack/alert-testing/npd-rule-tests.yaml
```diff
@@ -0,0 +1,174 @@
+# promtool unit test for Node Problem Detector alerts defined in
+#   ../npd-alerts-rules.yaml
+#
+# Run with:  ./extract-and-test.sh
```
DanielKChristou: The script is called `test.sh`. Same issue on line 7 below.

DanielKChristou:
Just something that I spotted while checking, this alert will fire fine in Prometheus, but there's an existing routing rule that ends up silencing it before it gets to Slack because its name overlaps with an existing filter:

https://github.com/hmcts/cnp-flux-config/blob/master/apps/monitoring/kube-prometheus-stack/kube-prometheus-stack.yaml#L101-L104

The alert itself is here:

https://github.com/hmcts/cnp-flux-config/blob/DTSPO-33407-npd-alerts-rules/apps/monitoring/kube-prometheus-stack/npd-alerts-rules.yaml#L48

Quickest fix might be to just rename it to something like NPDKubeletProblem so it doesn't overlap with that filter.

DanielKChristou (changes requested):

apps/monitoring/kube-prometheus-stack/alert-testing/test.sh
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env bash
+#
+# extract-and-test.sh
```
DanielKChristou: Same here

MartyFox:
> Just something that I spotted while checking, this alert will fire fine in Prometheus, but there's an existing routing rule that ends up silencing it before it gets to Slack because its name overlaps with an existing filter:
> 
> https://github.com/hmcts/cnp-flux-config/blob/master/apps/monitoring/kube-prometheus-stack/kube-prometheus-stack.yaml#L101-L104
> 
> The alert itself is here:
> 
> https://github.com/hmcts/cnp-flux-config/blob/DTSPO-33407-npd-alerts-rules/apps/monitoring/kube-prometheus-stack/npd-alerts-rules.yaml#L48
> 
> Quickest fix might be to just rename it to something like NPDKubeletProblem so it doesn't overlap with that filter.

I believe the `match_re` is match regex but the regex isn't using any expansion so its exact string, NPD** alerts wont match, those that match critical will get picked up and sent to critical slack channel.

DanielKChristou (approved):

PR #47008, opened 2026-08-21
fix(AMP-1018): scope apim-marketplace to sbox only

coling01:
## Summary
- Removes `apim-marketplace` HelmRelease from `cnp/base` (which applies to all CNP environments including AAT)
- Adds it directly to `cnp/sbox/base` so it only deploys to the sbox cluster
- Prevents Flux from conflicting with the service's existing CI-managed AAT deployment, which was failing with vault 403 errors

## Test plan
- [ ] AAT pod (`apim-marketplace-java`) should be removed and the CI-managed deployment restored
- [ ] Sbox pod should start cleanly with vault disabled (via existing `sbox.yaml` patch)


thomast1906 (dismissed):
should not be in apps/cnp/* - keep it your own apim folder - https://github.com/hmcts/cnp-flux-config/tree/fb0c2cd63078425288ebc785f00181af96fc8b88/apps/apim 

probably want to change naming to actual apim-marketplace, rather than apim

playfair0319-moj (dismissed):

willwatters (approved):

PR #47271, opened 2026-09-04
AMP-1071: mount the redis and session secrets for apim-marketplace-web in sandbox

coling01:
Mounts the Redis and session secrets for `apim-marketplace-web` in sandbox, so sessions survive being served by either cluster.

## Why

Each environment runs two AKS clusters (`cft-sbox-00-aks` and `cft-sbox-01-aks`) and flux deploys to both. The load balancer spreads requests across them, so with sessions held in each process's memory a signed-in user is signed out roughly half the time.

Measured on sandbox — twenty requests to `/info` returned ten from each of two pods, on different ReplicaSets running the same build:

```
10 apim-marketplace-web-nodejs-cb87d4455-85t5k
10 apim-marketplace-web-nodejs-844475b5b-j8rkw
```

A shared Redis with a shared `SESSION_SECRET` fixes it — both halves must match or the second process rejects the cookie.

## What this does

The vault key is `apim-sbox` with `excludeEnvironmentSuffix: true`, matching how `apim-marketplace` already reaches its Postgres secrets in this environment. `workloadClientID` is the sandbox identity `775648eb-…`.

No `REDIS_TLS` alias — the app defaults to TLS on port 6380, which is what the terraform publishes.

## The secrets already exist

Created by the sandbox parameterised pipeline (`sandbox-build.hmcts.net`, `HMCTS_j_to_z_Sandbox`), which only became possible for this repo with hmcts/web-api-marketplace#44:

```
marketplace-web-redis-host
marketplace-web-redis-port
marketplace-web-redis-access-key
marketplace-web-session-secret
```

The cache `apim-marketplace-web-redis-cache-sandbox` is provisioned and `Succeeded`, so mounting these will not give `Init:0/1`.

The application side is already merged and deployed (hmcts/web-api-marketplace#45). Until this merges it logs `No REDIS_HOST configured, holding sessions in memory for this process only` and behaves as before, so merging this is what switches the shared store on.

## Review changes

- Sandbox replica overrides removed at Tomas's request. That left `apim-marketplace/sbox.yaml` identical to master, so it is no longer in this PR.
- AAT overlay dropped — not needed yet, and its secrets do not exist. Sandbox uses `apim-sbox` in `DCD-CFT-Sandbox`; aat and demo use `apim-aat` and `apim-demo` in `DCD-CNP-DEV`, each needing its own terraform run first.
- On moving the block into `apim-marketplace-web.yaml`: with AAT dropped there is no longer any duplication, and that base file also covers `demo`, which has no Redis, so the mount would fail there.

This is now a single file, 15 lines.

### DIFF
--- a/apps/apim/apim-marketplace-web/sbox.yaml
+++ b/apps/apim/apim-marketplace-web/sbox.yaml
@@ -7,7 +7,6 @@ spec:
     global:
       enableKeyVaults: true
     nodejs:
-      replicas: 1
       ingressHost: apim-marketplace-web.sandbox.platform.hmcts.net
       workloadClientID: "775648eb-07db-4354-90f6-f0bca549341b"
       keyVaults:
--- a/apps/apim/apim-marketplace-web/sbox.yaml
+++ b/apps/apim/apim-marketplace-web/sbox.yaml
@@ -4,5 +4,21 @@ metadata:
   name: apim-marketplace-web
 spec:
   values:
+    global:
+      enableKeyVaults: true
     nodejs:
+      replicas: 1
       ingressHost: apim-marketplace-web.sandbox.platform.hmcts.net
+      workloadClientID: "775648eb-07db-4354-90f6-f0bca549341b"
+      keyVaults:
+        apim-sbox:
+          excludeEnvironmentSuffix: true
+          secrets:
+            - name: marketplace-web-redis-host
+              alias: REDIS_HOST
+            - name: marketplace-web-redis-port
+              alias: REDIS_PORT
+            - name: marketplace-web-redis-access-key
+              alias: REDIS_KEY
+            - name: marketplace-web-session-secret
+              alias: SESSION_SECRET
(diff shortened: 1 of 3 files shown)


apps/apim/apim-marketplace-web/sbox.yaml
```diff
@@ -4,5 +4,21 @@ metadata:
   name: apim-marketplace-web
 spec:
   values:
+    global:
+      enableKeyVaults: true
     nodejs:
+      replicas: 1
```
thomast1906: dont just have 1 replica

thomast1906 (changes requested):
remove replicas: 1 - default is 2 per cluster

apps/apim/apim-marketplace-web/sbox.yaml
coling01: it just sandbox? and 2 clusters ...

thomast1906 (approved):

PR #47308, opened 2026-09-07
Increase artifactory Web UI artefact upload limit to allow caps artefact to be uploaded (210mb)

dawidstrozak:
### Jira link

https://tools.hmcts.net/jira/browse/DTSPO-33744

### Change description

Currently limit is 100mb so UI won't allow for anything larger to be uploaded.
Tried uploading using curl in two different ways but either end up with a 0 byte file or get 502 response, not sure why.
It is better to be able to use UI anyway, it will simplify the recovery playbook.

### Testing done

- 

### Security Vulnerability Assessment ###

**CVE Suppression:** Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?
  * [ ] Yes
  * [x] No

### Checklist


endakelly (approved):
Does artifactory do a dynamic reload without restarting the pods?

dawidstrozak:
> Does artifactory do a dynamic reload without restarting the pods?

That is a good question, I did change some stuff in the past when I was patching it but I don't remember if it dynamically reloaded. Will find out I suppose.
EDIT: yeah looks like it does not, it needs a restart

PR #47326, opened 2026-09-07
Revert "adding ingressHost to resolve dns issue"

tmckillop95:
Reverts hmcts/cnp-flux-config#47315


endakelly (approved):
Those failures just need re-runs. Sometimes downloads time out

PR #47491, opened 2026-09-14
Create the deed-poll namespace in cft-aat and cft-preview

NewmanJustice:
Namespace-only onboarding for a new product, `deed-poll`. Same shape as `apps/coe`, `apps/crumble` and `apps/halo`: four files under `apps/deed-poll/`, one appended line in each of the two cluster kustomizations, and a CODEOWNERS entry.

**It deploys nothing.** No `HelmRelease`, no `ImageRepository`/`ImagePolicy`, no `automation/` entry. The only runtime effect is that the namespace and its standard RBAC/alerting objects exist. HelmReleases follow in a later PR, once the charts are published.

**7 files, +34, zero deletions** — no existing line anywhere was modified or reordered, only appended to.

`TEAM_AAD_GROUP_ID` is `446b9d16-60ae-41c0-8678-8430cbbd1c27`, resolved with `az ad group show` against the live `DTS Deed Poll` group (created today by azure-access#8110), not hand-written. `deed-poll` is registered in `cnp-jenkins-config` `team-config.yml` as of #1339, and the namespace and Slack channel here match that entry.

Validation, all green and identical to the unmodified tree apart from our additions:

- `yamllint . -c .yamllint.yaml` — exit 0, sole output the pre-existing `.github/workflows/stale.yml` truthy warning
- `kustomize build clusters/aat/base` and `clusters/preview/base` — exit 0
- `bash tests/image-policy-exists.sh` — exit 0, no `deed-poll` lines (nothing to police yet)
- `kustomize build apps/deed-poll/aat/base` — **7 objects**: `Namespace`, `ServiceAccount deed-poll`, `ServiceAccount tests-service-account`, `Role` + `RoleBinding nonprod-team-permissions`, `Alert team-alerts`, `Provider slack`
- `kustomize build apps/deed-poll/preview/base` — **5 objects**, the same minus Role/RoleBinding (no `nonprod-role.yaml` in preview)
- `bash tests/code-owner-check.sh` — exit 0, both assertions ran

Note the CODEOWNERS entry does not remove the need for PlatOps review here — the later kustomization rules out-rank it. It pays off in the follow-on PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


NewmanJustice:
Pushed `1d33b37` to fix the CODEOWNERS validation failure on line 178.

**What was wrong:** the error lists three conditions, and two were already satisfied —

- the team **does** exist, and GitHub resolves the slug (`adoption-deedpoll`, display name `Adoption-Deedpoll`)
- its privacy is `closed`, i.e. visible within the org
- ❌ **it has access to no repositories** — `repos_count` is 0, so it has never been granted write anywhere, this repo included

GitHub won't accept a CODEOWNER that can't write to the repo it owns paths in, so the slug was never the problem.

**Fix:** `apps/deed-poll/` is now owned by `@hmcts/reform-adoption`, which already owns `apps/adoption/` in this file and has 13 repositories and 15 members. That's accurate rather than a workaround — deed poll sits with the Adoption team today.

`bash tests/code-owner-check.sh` passes 2/2; the two PlatOps kustomization rules are untouched.

**Follow-up, not needed for this PR:** the intended owner is still `adoption-deedpoll`, and switching is a one-line change once someone with org admin grants that team write access here. Doing that first would have meant a second PlatOps request stacked on this review, and this PR is the last gate before a first deed-poll deployment — the AAT infrastructure applied successfully earlier today.

thomast1906 (approved):
PR fine , but comment: Worth just adding it now to the adoption namespace?

PR #47540, opened 2026-09-16
Add a LiteLLM gateway proxy to dtsse in AAT

linusnorton:
[VIBE-574](https://hmcts.atlassian.net/browse/VIBE-574)

An nginx reverse proxy in the `dtsse` namespace on `litellm.aat.platform.hmcts.net`, forwarding to a LiteLLM gateway running on ECS in the MoJ Modernisation Platform.

Engineers use that gateway to run Claude Code against Bedrock. Fronting it from AAT means the hostname resolves on the private zone only, and outbound traffic leaves via the hub firewall's fixed addresses (`20.50.108.242`, `20.50.109.148`), so the gateway's AWS WAF can be narrowed to those and taken off the public internet.

Deployed to both AAT clusters via `aat/base` and load balanced by the backend application gateway.

- Plain manifests rather than a HelmRelease: no image to build and no chart to publish. Same approach as `apps/money-claims/cmc-mailhog`.
- No TLS router annotation. The application gateway terminates TLS and connects to traefik over HTTP on port 80, so a TLS-only router would leave its requests and health probe unmatched.
- `proxy_buffering off` and 900s timeouts to match the upstream load balancer idle timeout.
- The upstream is resolved per request through a variable, as its addresses change and nginx would otherwise pin them at startup.
- Probes use a local `/healthz`, so an upstream outage returns errors rather than restarting pods.
- Image is `imported/nginx`, an existing ACR cache rule in hmcts/acr-base-importer.

Depends on:

- hmcts/azure-platform-terraform#3154 for the listener, backend pool and health probe.
- hmcts/azure-private-dns#1071 for the record.
- AAT firewall egress to `claude-gateway.hmcts-development.modernisation-platform.service.justice.gov.uk`, which is not yet confirmed.

[VIBE-574]: https://hmcts.atlassian.net/browse/VIBE-574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


thomast1906 (changes requested):
lets discuss this

thomast1906:
^ we wont deploy it into CFT - we'd deploy the proxy piece into AI integration gateway

PR #47543, opened 2026-09-16
Give deed-poll a workload identity, so its pods can read their own vault

NewmanJustice:
`deed-poll-admin`'s first deploy to AAT never leaves `Init:0/1`, and `helm upgrade --wait` times out after 20 minutes:

```
MountVolume.SetUp failed for volume "vault-deed-poll":
failed to get objectType:secret, objectName:app-insights-connection-string
nmi response failed with status code: 404
getting assigned identities for pod deed-poll/deed-poll-admin-staging-nodejs-… failed after 16 attempts
```

The image, the chart and the vault are all fine — `deed-poll-aat` holds the secret and the namespace is live. The pod simply has no identity to exchange for a token.

### What's missing

`apps/deed-poll/` has never included the workload-identity base, so nothing creates the federated credential that binds the identity to the namespace service account — and the service account carries no client-id annotation to present. Both halves are here, and both match `apps/pcs`:

| Change | Effect |
|---|---|
| `apps/deed-poll/base/kustomization.yaml` gains `../../base/workload-identity` | creates the `FederatedIdentityCredential` for `system:serviceaccount:deed-poll:deed-poll` |
| new `apps/deed-poll/serviceaccount/aat.yaml` | annotates the service account with `deed-poll-aat-mi`'s client id |
| `apps/deed-poll/aat/base/kustomization.yaml` gains a `patches:` entry | applies that patch in AAT |
| `apps/deed-poll/preview/base/kustomization.yaml` gains both | same, on the preview cluster, where PR builds deploy — see below |

### Why no other change is needed

The substitution was already in place — `apps/deed-poll/base/kustomize.yaml` sets `WI_NAME: "deed-poll"`, so the variable existed for a resource that was never pulled in. deed-poll's substitution block is otherwise **identical** to pcs's (`NAMESPACE`, `WI_NAME`, `TEAM_NOTIFICATION_CHANNEL`, `TEAM_AAD_GROUP_ID`), so `WI_ENVIRONMENT`, `WI_CLUSTER` and `ISSUER_URL` come from cluster-level config as they do for pcs.

There is exactly one service account to patch, so the patch merges rather than competing: `apps/base/kustomization.yaml` includes `service-account.yaml`, and `apps/base/workload-identity/kustomization.yaml` deliberately does **not** list its own `service-account.yaml`.

### ASO will not fight Terraform — and an earlier revision of this PR asked the wrong question

An earlier revision asked a reviewer to confirm whether ASO would adopt or collide with a pre-existing `deed-poll-aat-mi`. **That resolves from the manifests in this repo, so it does not need reviewer time:**

- **Terraform creates the identity, not ASO.** `cnp-module-key-vault/managed-identity.tf:6` creates `${var.product}-${local.env}-mi` in `managed-identities-${local.env}-rg`, and its `implicit_managed_identity_access_policy` grants that identity `Get`/`List` on the vault's secrets. `hmcts/deed-poll-infra` consumes the module.
- **`apps/base/workload-identity/workload-identity-ua-identity.yaml` carries `serviceoperator.azure.com/reconcile-policy: skip`**, so ASO takes no action on the identity in Azure at all. The CR exists so that `FederatedIdentityCredential` can name it as `owner`.
- `workload-identity-rg.yaml` is `detach-on-delete`, and `managed-identities-aat-rg` already exists, so ASO adopts it rather than recreating it.

So what this PR adds in Azure is the federated credential, and nothing else.

Verified live rather than assumed:

| | |
|---|---|
| `deed-poll-aat-mi` | exists — `managed-identities-aat-rg`, subscription `1c4f0704-…` |
| its client id | `b116da08-b7ed-4059-b655-08670ea1ed15`, the value in `serviceaccount/aat.yaml` |
| its principal `687fa087-…` | holds `Get,List` on `deed-poll-aat`'s secrets |

### The companion change, and merge order

[`hmcts/deed-poll-admin#103`](https://github.com/hmcts/deed-poll-admin/pull/103) sets `useWorkloadIdentity` and `saEnabled` on the chart, drops `aadIdentityName`, and takes `nodejs` 3.2.1 → 3.3.1. **Both PRs are needed and neither works alone:**

- without `#103`, chart-library emits `aadpodidbinding` instead of `azure.workload.identity/use: "true"`, so the pod never requests workload identity;
- also without `#103`, chart-library gives the pod no `serviceAccountName` at all, so it runs as `default` — and this PR's federated credential binds `deed-poll`. Annotating a service account no pod presents produces the same NMI 404 with a different-looking explanation;
- without this PR, `#103` asks for a workload identity with no federated credential behind it.

### Preview is covered too, because PR builds deploy one

Added after reading `deed-poll-admin`'s `PR-103/1` log, which settled a question this PR had guessed at. **PR builds do deploy a preview:** `helm upgrade deed-poll-admin-pr-103 … --namespace deed-poll`, under the lock `pr-103-deed-poll-admin-preview-deploy`, and the pod sat at `Init:0/1` for 20 minutes with the same NMI 404 before the stage aborted.

Preview is a **separate cluster** — `clusters/preview/base/kustomization.yaml:67` and `clusters/aat/base/kustomization.yaml:70` each include `apps/deed-poll/base/kustomize.yaml` — and `apps/deed-poll/preview/base` resolves `../../../base`, the shared `apps/base`, not `apps/deed-poll/base`. So the AAT change above does not reach it, and without the preview change **every PR build would burn 40 minutes and go red even once AAT works.**

It reuses the same identity rather than needing a second, which is why a preview overlay patches a file named `aat.yaml`:

```
apps/flux-system/preview/00/kustomize.yaml   ENVIRONMENT: "preview"  WI_ENVIRONMENT: "aat"
apps/flux-system/aat/00/kustomize.yaml       ENVIRONMENT: "aat"      WI_ENVIRONMENT: "aat"
```

`WI_ENVIRONMENT` is `aat` on both clusters, so both own `deed-poll-aat-mi`; `WI_CLUSTER` is `${ENVIRONMENT}-${CLUSTER}`, so each cluster gets its own credential on that one identity — `deed-poll-preview-00-fic` alongside `deed-poll-aat-00-fic`. `apps/pcs/preview/base/kustomization.yaml` does exactly this.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


NewmanJustice:
Adding two findings from digging into why the deploy fails, both of which change how this PR should be read. **Please don't merge it as a standalone fix** — it is one of three pieces.

### 1. `deed-poll-aat-mi` already exists, created by Terraform

`cnp-module-key-vault/managed-identity.tf:6` creates `${var.product}-${local.env}-mi` in `managed-identities-${local.env}-rg`, and `hmcts/deed-poll-infra` consumes that module. So this PR's ASO `UserAssignedIdentity` and `ResourceGroup` will be reconciling resources Terraform already owns.

**I believe that's fine, because pcs is in exactly the same position** — `pcs-shared-infrastructure/key-vault.tf` also uses `cnp-module-key-vault`, and `apps/pcs/base/kustomization.yaml` also lists `../../base/workload-identity`. But you'd know better than me whether ASO adopts cleanly here or whether there's a preferred way to express it, so I'd rather ask than assume.

### 2. The chart was asking for pod identity, not workload identity

`deed-poll-admin`'s chart set `aadIdentityName: deed-poll`, which is why the error mentions NMI at all — chart-library only emits `azure.workload.identity/use: "true"` when `useWorkloadIdentity` is set, and `_labels.tpl` emits `aadpodidbinding` otherwise. So workload identity was never actually being requested by the pod.

There was a second gap in the same place: without `saEnabled`, chart-library gives the pod **no** `serviceAccountName`, so it runs as `default` — and the federated credential this PR creates binds `system:serviceaccount:deed-poll:deed-poll`. Annotating that service account while the pod runs as `default` would federate an identity no pod ever presents.

Both are fixed in [`hmcts/deed-poll-admin#103`](https://github.com/hmcts/deed-poll-admin/pull/103), which also takes `nodejs` 3.2.1 → 3.3.1 (the version `check-deprecated-charts.sh` already demands, and the one carrying chart-library 2.3.1 where those values live).

### What's still outstanding here

The service-account annotation, which is what actually completes it:

- `apps/deed-poll/serviceaccount/aat.yaml` — a `ServiceAccount` patch with `azure.workload.identity/client-id`
- a `patches:` entry for it in `apps/deed-poll/aat/base/kustomization.yaml`

exactly as `apps/pcs` has. That needs `deed-poll-aat-mi`'s client id, which is a literal GUID in pcs's equivalent file — I don't have read access to the subscription holding `managed-identities-aat-rg`, so I'll add it as a follow-up commit here once I have the value rather than guess at it.

Happy to fold everything into this PR instead if you'd prefer one change over two.

NewmanJustice:
Folded the follow-up in, so this is now the whole flux-side change rather than one of two — `4d5ca9e` adds `apps/deed-poll/serviceaccount/aat.yaml` and its `patches:` entry. The description above has been rewritten to match; **please read it rather than my comment above, which is now partly superseded.**

**I also answered my own question from that comment, and the answer is better than "pcs does it too".** `apps/base/workload-identity/workload-identity-ua-identity.yaml` carries `serviceoperator.azure.com/reconcile-policy: skip`, so ASO takes no action on the `UserAssignedIdentity` in Azure — the CR exists only so the `FederatedIdentityCredential` can name it as `owner`. Terraform creates the identity (`cnp-module-key-vault/managed-identity.tf:6`) and ASO leaves it alone by design. The `ResourceGroup` is `detach-on-delete` and `managed-identities-aat-rg` already exists, so it adopts. **So there is nothing for you to adjudicate there** — the only thing this PR creates in Azure is the federated credential. Sorry for the noise.

The client id came from `az graph query` against `managed-identities-aat-rg` rather than a guess, and I checked the identity's principal already holds `Get`/`List` on `deed-poll-aat`'s secrets — so the vault side needs nothing.

Two things still worth your eye, and neither is the one I asked about:

1. **Merge order.** `hmcts/deed-poll-admin#103` has to land too — without it the pod runs as `default` rather than `deed-poll` and never presents the identity this PR federates. Either order is fine; both before the next deploy attempt.
2. **Preview is deliberately not covered here.** `apps/deed-poll/preview/base/kustomization.yaml` points at the shared `apps/base`, not at `apps/deed-poll/base`, so it gets neither the workload-identity base nor this patch. `apps/pcs` includes both in preview and shares the AAT identity. I've left it out to keep this scoped to the AAT failure — say if you'd rather it were done in one go.

adusumillipraveen (dismissed):

NewmanJustice:
`ac5985b` adds the same wiring to the preview cluster. **This corrects my last comment, where I said preview was deliberately out of scope — that was based on deed-poll never having deployed a preview, and it turned out to be wrong the same day.**

`deed-poll-admin`'s `PR-103/1` log shows PR builds *do* deploy a preview: `helm upgrade deed-poll-admin-pr-103 … --namespace deed-poll` under the lock `pr-103-deed-poll-admin-preview-deploy`, pod at `Init:0/1` for 20 minutes with the same NMI 404, stage aborted after 46 minutes. Since preview is a separate cluster and `apps/deed-poll/preview/base` resolves the shared `apps/base` rather than `apps/deed-poll/base`, the AAT change alone would have left **every PR build in that repo red**.

No new identity or GUID: `WI_ENVIRONMENT` is `aat` on both the preview and aat clusters, so both own `deed-poll-aat-mi`, and `WI_CLUSTER` (`${ENVIRONMENT}-${CLUSTER}`) gives each its own federated credential on it — `deed-poll-preview-00-fic` alongside `deed-poll-aat-00-fic`. That is why a preview overlay patches a file called `aat.yaml`, matching `apps/pcs/preview/base`.

Rendering `apps/deed-poll/preview/base` before and after adds the client-id annotation and the three workload-identity objects, and nothing else.

`hmcts/deed-poll-admin#103` is now **merged** (`d86fff2`), so this PR is the only remaining piece.

hfilmore (approved):

adusumillipraveen (approved):

